PR #4, opened 2023-03-02
add Khadas Edge 2

rpardini:
Picked from https://github.com/armbian/build/pull/4805 I had sent before and was ignored. I've report this works on rk3.4, and picked cleanly on top of this new rk3.6 branch. Probably needs some cleanup. Commits are straight from Khadas, except when indicated by `[COMMENT]`


rpardini:
I didn't squash in the hopes that Khadas has updates we might wanna pick, but really we should squash I think. too many back-and-forths to be useful

rpardini:
Might be something more to pick at https://github.com/khadas/linux/commits/khadas-edge2-android12

amazingfate:
> Might be something more to pick at https://github.com/khadas/linux/commits/khadas-edge2-android12

I will take a look at this. We can merge all the dts commit together so that we can have a clean commit history.

150balbes:
> I will take a look at this.

don't rush to accept it, this needs to be checked

PR #10, opened 2023-03-18
cma: Enable reserved-memory on orangepi5

neotty:
Fix the cma_alloc: alloc failed problem.
image


amazingfate (approved):

amazingfate:
Just curious, which component uses cma since there is no hdmirx on opi5?

neotty:
This error occurred when I scrcpy to the android system running in docker~

PR #11, opened 2023-03-22
Armbian 9tripod / Indienova patchset

stvhay:


arch/arm64/boot/dts/rockchip/rk3588-linux.dtsi
amazingfate, line 1: I don't want to change this file because it's from the upstream SDK. You can delete this node in the board dts like other board: https://github.com/armbian/linux-rockchip/blob/rockchip-5.10/arch/arm64/boot/dts/rockchip/rk3588-rock-5b.dts#L29

arch/arm64/boot/dts/rockchip/rk3588s.dtsi
amazingfate, line 1: This is not a board level change and will affect other boards. You can move this to another pr so that we can discuess this change there.

drivers/net/wireless/rockchip_wlan/rtl8821cs/core/monitor/rtw_radiotap.c
amazingfate, line 1: There are many changes like this, I don't think it is necessary.

amazingfate (changes requested):

arch/arm64/boot/dts/rockchip/rk3588-linux.dtsi
stvhay: That's reasonable to me

arch/arm64/boot/dts/rockchip/rk3588s.dtsi
stvhay: Sure

drivers/net/wireless/rockchip_wlan/rtl8821cs/core/monitor/rtw_radiotap.c
stvhay: This commit replaces an older version of the driver with newer version. Wouldn't we want to sync with that version vs creating a third custom version of the code where possible?

rpardini:
This `rockchip_wlan` driver stuff is puzzling.

arch/arm64/boot/dts/rockchip/rk3588-linux.dtsi
rpardini: I had the same on the Mekotronics. Using the trick amazingfate mentions (delete the node at the board DTS) works.

arch/arm64/boot/dts/rockchip/rk3588s.dtsi
rpardini: Agreed, no changing the common dtsi is needed.

PR #12, opened 2023-04-01
add Mekotronics-R58

rpardini:
DTS's, a hack for `stmmac` LEDs, and that pesky `chosen` `bootargs` root reference that had popped up before but lost in time. DT's and stmmac hack provided by @monkaBlyat.


rpardini:
Force pushed, updated some Mekotronic product names in the DT.

arch/arm64/boot/dts/rockchip/rk3588-linux.dtsi
amazingfate, line 1: We don't have to change this upstream device tree. We can treat it downstream like rock5b: https://github.com/armbian/linux-rockchip/blob/rockchip-5.10/arch/arm64/boot/dts/rockchip/rk3588-rock-5b.dts#L29

amazingfate (changes requested):

arch/arm64/boot/dts/rockchip/rk3588-linux.dtsi
rpardini: Agreed! Will do.

rpardini: Done...

amazingfate (approved):

PR #14, opened 2023-04-14
mekotronics: drivers: rk_hdmirx: shut up spammy logs

rpardini:
Couldn't find a better way.
If the HDMI-in is not plugged in, a lot of dmesg spam occurs - it can't lock the signal.
This is not only cosmetic, having an overflowing dmesg buffer can hamper the execution of `sbc-bench`, `armbianmonitor` etc.
This is solving it with a hammer, if anyone has a better idea?

### DIFF
--- a/drivers/media/platform/rockchip/hdmirx/rk_hdmirx.c
+++ b/drivers/media/platform/rockchip/hdmirx/rk_hdmirx.c
@@ -1516,10 +1516,10 @@ static int hdmirx_wait_lock_and_get_timing(struct rk_hdmirx_dev *hdmirx_dev)
 	}
 
 	if (i == WAIT_SIGNAL_LOCK_TIME) {
-		v4l2_err(v4l2_dev, "%s signal not lock, tmds_clk_ratio:%d\n",
-				__func__, hdmirx_dev->tmds_clk_ratio);
-		v4l2_err(v4l2_dev, "%s mu_st:%#x, scdc_st:%#x, dma_st10:%#x\n",
-				__func__, mu_status, scdc_status, dma_st10);
+		// spammy // v4l2_err(v4l2_dev, "%s signal not lock, tmds_clk_ratio:%d\n",
+		// spammy // 		__func__, hdmirx_dev->tmds_clk_ratio);
+		// spammy // v4l2_err(v4l2_dev, "%s mu_st:%#x, scdc_st:%#x, dma_st10:%#x\n",
+		// spammy // 		__func__, mu_status, scdc_status, dma_st10);
 
 		return -1;
 	}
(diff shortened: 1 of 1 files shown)


rpardini:
This is indeed a terrible way. Removing the logs would cripple useful logging for _actually working_ hdmirx on other boards.
Problem probably lies in `det` gpio on the R58's DTS. Since I've not the schematics for the boards, I could only speculate.

From chat history, thanks @amazingfate 

---------

amazingfate — Yesterday at 2:24 PM
I don't have hdmi rx spamming logs when the hdmi input source is not plugged on rock5b and another rk3588 board. Maybe we can investigate further.

rpardini — Yesterday at 2:27 PM
it is probably related to some missing pins in the Meko's DTS, but beyond my skillset
some det pin or such, which would signal "hey a thing is actually plugged in"

amazingfate — Yesterday at 2:31 PM
And maybe some interference. I got constantly hdmirx reset when both connected to hdmi input and wireless: https://forum.radxa.com/t/hdmi-in-frequently-reset/15760

rpardini — Yesterday at 2:34 PM
Yeah interference would explain errors if I had something pluggged in and it did not work... 
I haven't plugged in anything yet
I was thinking maybe something here?
- https://github.com/armbian/linux-rockchip/blob/rockchip-5.10/arch/arm64/boot/dts/rockchip/rk3588-rock-5b.dts#L634-L642
vs
- https://github.com/armbian/linux-rockchip/blob/rockchip-5.10/arch/arm64/boot/dts/rockchip/rk3588-blueberry-edge-v12.dtsi#L356

amazingfate — Yesterday at 3:54 PM
I will spend sometime to see when these logs will show up.

rpardini — Yesterday at 4:17 PM
Yeah don't waste much time; having non-flooded dmesg is pretty important but if all else fails I might just disable the hdmirx and add an overlay to enable it.
And super thanks for the help.

--------

amazingfate — Today at 3:47 AM
the hdmirx driver will check the gpio to see if hdmi input is plugged: https://github.com/armbian/linux-rockchip/blob/rockchip-5.10/drivers/media/platform/rockchip/hdmirx/rk_hdmirx.c#L507
You can check if hdmirx_det_gpio is correctly configed in the device tree. 

rpardini — Today at 3:59 AM
Yep, that's gotta be it. Thanks.
- Rock-5b: `hdmirx-det-gpios = <&gpio1 RK_PC6 GPIO_ACTIVE_LOW>;` 
- Mekotronics: `hdmirx-det-gpios = <&gpio1 29 GPIO_ACTIVE_LOW>;   //gpio1_d5`
- Rock-5b: matches schematics https://dl.radxa.com/rock5/5b/docs/hw/radxa_rock_5b_v142_sch.pdf page 8
- Mekotronics: _which schematics?_
So yeah I'm gonna disable that hdmirx and add overlay for now.

PR #15, opened 2023-04-15
arm64: dts: Add device trees for NanoPi R6S and R6C

efectn:


rpardini (approved):
Looks good. Doesn't touch any other boards.

PR #16, opened 2023-04-19
arm64: dts: Add device tree for Hinlink H88K

amazingfate:
Hinlink H88K is a rk3588 board with full IO: https://doc.hinlink.cn/
Test images: https://github.com/amazingfate/armbian-h88k-images


rpardini (approved):
clean DTS, no other boards touched, go

PR #17, opened 2023-04-21
arm64: dts: enable sdhci node in OPi5 device tree

efectn:


igorpecovnik:
Yes, I am fine with such solution.

amazingfate:
What's the issue we are going to solve? Why the opi5b device tree has to get deleted?

igorpecovnik:
> Why the opi5b device tree has to get deleted?

It seems no need to have a dedicated device tree for b model. Or?

efectn:
> What's the issue we are going to solve? Why the opi5b device tree has to get deleted?

If we enable sdhci, we won't need device tree for 5b anymore. There are no any other differences between 5 and 5b. So it's overkill to provide new devicetree

amazingfate:
You may need to add `wireless_bluetooth` and `wireless_wireless` and their pinctl to the device tree of opi5 if you want to support opi5b using the same dts.

efectn:
> You may need to add `wireless_bluetooth` and `wireless_wireless` and their pinctl to the device tree of opi5 if you want to support opi5b using the same dts.

Yeah saw it. Will do it when i go to the home

efectn:
Looks like it's not necessary https://forum.armbian.com/topic/27986-support-to-orange-pi-5b-with-ap6275p-wifi/

efectn:
Looks like it's not necessary https://forum.armbian.com/topic/27986-support-to-orange-pi-5b-with-ap6275p-wifi/

But can create an overlay to disable sfc etc.

amazingfate:
> Looks like it's not necessary https://forum.armbian.com/topic/27986-support-to-orange-pi-5b-with-ap6275p-wifi/
> 
> But can create an overlay to disable sfc etc.

That overlay is not is this repo, you can add it back.

efectn:
Just enabled wlan and bluetooth nodes instead of adding an overlay

amazingfate (approved):

PR #18, opened 2023-04-24
arm64:dts: Add pwm-fan node to rock5b

amazingfate:
Let kernel pwm-fan driver deal with the cooling control. Although different fan/case combinations need different cooling step control, I think we can add a default one works for the radxa official fan.
Tested with the radxa official fan inside my 3D printed case. The lowest pwm vaule to make the fan spin is 72. I set it as the start level because it will keep the soc temp in a stable range. In my case when the fan is running at the lowest speed, the noise is not that big and the soc temp is about 42.5℃ when there is no system load.
Users can wirte overlays to config their owm fan and case. Here is the overlay I used before: https://github.com/amazingfate/radxa-rock5b-overlays/blob/main/pwm-fan.dts. This commit is also based on this overlay.

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3588-rock-5b.dts
+++ b/arch/arm64/boot/dts/rockchip/rk3588-rock-5b.dts
@@ -256,6 +256,13 @@
 			sound-dai = <&hdmiin_dc>;
 		};
 	};
+
+	fan0: pwm-fan {
+		compatible = "pwm-fan";
+		#cooling-cells = <2>;
+		cooling-levels = <72 94 117 139 162 184 207 229 255>;
+		pwms = <&pwm1 0 10000 0>;
+	};
 };
 
 &av1d {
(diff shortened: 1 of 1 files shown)


amazingfate:
@ginkage What do you think of this? Do I have to change the default pwm config like `cooling-levels`?

ginkage:
I don't have a strong opinion about the curve. The change itself is very welcome indeed.

amazingfate:
Users can config `cooling-levels` by an overlay like this:
```
/dts-v1/;
/plugin/;
/ {
        fragment@0 {
                target = <&fan0>;
                __overlay__ {
                        cooling-levels = <0 115 160 205 255>;
                };
        };
}
```

ginkage:
OK, I've played with a number of curves now, and now I have a suggestion to add just one more control point at 51 (or maybe even 50) degrees: I see you're doing a ~12 degree step, but I prefer a little bit more granularity on the left side of the graph, so it's nice to break that interval in two. That way I can implement the curve that's better suited for my setup, e.g.: <72 120 163 189 226 255>

amazingfate:
@ginkage I changed the temp step to 5 degrees.

ginkage:
Yeah, that works too. :)

ginkage (approved):

rpardini (approved):
LGTM. Pinging @lanefu and @Tonymac32 for awareness since "can rock-5b pwm the fan" had come up an year or so ago...

q4a:
@amazingfate Hi. Should someone add it to https://github.com/armbian/build/tree/main/patch/kernel/archive/rockchip-6.3 ?

rpardini:
@q4a that's an unrelated kernel... this is 5.10 legacy, you just pointed to mainline/edge.

q4a:
~@rpardini I pointed to mainline/edge because I'm interested if something similar should be added to it (not to 5.10 legacy)?~

UPD: sorry, I thought that this is not merged yet: https://patchwork.kernel.org/project/linux-rockchip/list/?series=736913

PR #27, opened 2023-05-22
Revert "khadas: sound/soc"

amazingfate:
This reverts commit 0f8095979ba1a6ab9e5b0dc1050f0dda3e77dcc9.
Khadas add `spk_ctl_gpio`, `hp_det_gpio` to the device tree node if es8316,but other boards such as rock5b and hinlink h88k don't have these properties. This will casuse kernel error on rock5b.


rpardini (approved):
Yes, lets revert. Thanks!

PR #32, opened 2023-05-26
arm64: dts: Add device tree for NanoPC T6

MollySophia:
Imported from https://github.com/friendlyarm/kernel-rockchip/blob/nanopi5-v5.10.y_opt/arch/arm64/boot/dts/rockchip/rk3588-nanopi6-rev01.dts

Tested on my NanoPC T6, and it works as-is


amazingfate (approved):
LGTM, and safe to merge.

igorpecovnik:
Welcome to add CSC build config here https://github.com/armbian/build/tree/main/config/boards

PR #35, opened 2023-06-03
arm64: dts: opi5: fix cursor and ap6275p

Joshua-Riek:
1. Fix WiFi and Bluetooth when using the ap6275p on the Orange Pi 5 / 5B.
2. Fix cursor issues (flickering, disappearing, etc.), similar to 4e169d4d1a296f9e85a53376df381b0f45447894 for the Rock 5B.


efectn (approved):
I don't have 5b to test it but seems fine

PR #38, opened 2023-06-05
arm64: dts: add opi5 plus device tree

Joshua-Riek:
Here is the initial device tree for the Orange Pi 5 Plus, extracted from Xunlong https://github.com/orangepi-xunlong/linux-orangepi/commit/7e6c3163aa7e58b19730aa2aa259f1bb957cbca0. I received my Orange Pi 5 Plus earlier today, and the device tree works as expected. Overlays will need to be added later.


amazingfate (approved):

ggg17226:
tested boot from emmc(jammy/cli), both 2 eth ports and nvme works fine.

PR #40, opened 2023-06-06
arm64: dts: add opi5b device tree again

Joshua-Riek:
The Orange Pi 5B eMMC conflicts with the SPI flash on the Orange Pi 5. So I propose adding the Orange Pi 5B device tree again to allow the eMMC to be used.

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3588s-orangepi-5b.dts
+++ b/arch/arm64/boot/dts/rockchip/rk3588s-orangepi-5b.dts
@@ -13,6 +13,10 @@
 	compatible = "rockchip,rk3588s-orangepi-5", "rockchip,rk3588";
 };
 
+&sdhci {
+	status = "okay";
+}
+
 &sfc {
 	status = "disabled";
 };
--- a/arch/arm64/boot/dts/rockchip/rk3588s-orangepi-5b.dts
+++ b/arch/arm64/boot/dts/rockchip/rk3588s-orangepi-5b.dts
@@ -15,7 +15,7 @@
 
 &sdhci {
 	status = "okay";
-}
+};
 
 &sfc {
 	status = "disabled";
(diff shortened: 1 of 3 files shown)


efectn (commented):
What about to include 5's devicetree and make the needed changes? Seems cleaner to me

Joshua-Riek:
That is a good idea since they are so similar. I'll work on the change later today.

Joshua-Riek:
Changes pushed, looks so much cleaner now.

efectn:
> Changes pushed, looks so much cleaner now.

Can you also disable sdhci node from arch/arm64/boot/dts/rockchip/rk3588s-orangepi.dtsi and enable it on 5b's tree

efectn (dismissed):

arch/arm64/boot/dts/rockchip/rk3588s-orangepi-5b.dts
```diff
@@ -0,0 +1,22 @@
...
+#include "rk3588s-orangepi-5.dts"
+
+/ {
+	model = "Orange Pi 5B";
+	compatible = "rockchip,rk3588s-orangepi-5", "rockchip,rk3588";
+};
+
+&sdhci {
+	status = "okay";
+}
```
amazingfate:
```
Error: arch/arm64/boot/dts/rockchip/rk3588s-orangepi-5b.dts:20.1-5 syntax error

[🐳|🔨]   FATAL ERROR: Unable to parse input tree
```
Missing `;`

amazingfate (changes requested):

arch/arm64/boot/dts/rockchip/rk3588s-orangepi-5b.dts
Joshua-Riek: Fixed!

amazingfate (approved):

PR #43, opened 2023-06-07
dts: rock5a: Update soc_ thermal node, providing more temperature opt…

FantasyGmm:
…ions
Referenced the device tree of Rock5b and tested its functionality under Armbian Ubuntu Jammy server

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3588s-rock-5a.dts
+++ b/arch/arm64/boot/dts/rockchip/rk3588s-rock-5a.dts
@@ -28,7 +28,7 @@
 	fan0: pwm-fan {
 		compatible = "pwm-fan";
 		#cooling-cells = <2>;
-		cooling-levels = <0 64 128 192 255>;
+		cooling-levels = <100 160 190 200 215 235 255>;
 		pwms = <&pwm3 0 10000 0>;
 	};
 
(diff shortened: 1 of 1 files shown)


arch/arm64/boot/dts/rockchip/rk3588s-rock-5a.dts
```diff
@@ -703,23 +703,74 @@
 };
 
 &threshold {
-	temperature = <60000>;
```
ginkage, line 706: Do we actually need this, if the reference at L719 has been removed?

FantasyGmm: The device tree of r3588s.dtsi still has trip0 1 and will be applied to the system

FantasyGmm: If I have time tomorrow, I will try deleting a node to see if it's possible

FantasyGmm: `threshold = "/thermal-zones/soc-thermal/trips/trip-point-0";`
```
root@rock-5a:~# cat /sys/class/thermal/thermal_zone0/trip_point_0_temp
60000
```
I checked the current 5A system device tree, but it has been applied,I haven't modified the device tree for this system yet

arch/arm64/boot/dts/rockchip/rk3588s-rock-5a.dts
```diff
@@ -28,7 +28,7 @@
 	fan0: pwm-fan {
 		compatible = "pwm-fan";
 		#cooling-cells = <2>;
-		cooling-levels = <0 64 128 192 255>;
+		cooling-levels = <100 160 190 200 215 235 255>;
```
amazingfate, line 31: My fan from radxa starts spinning from pwm value 135, does your fan starts at value 100?

FantasyGmm: > My fan from radxa starts spinning from pwm value 135, does your fan starts at value 100?
I have replaced the fan with a new one, and it will only start running at around 160. However, after the temperature drops to 100, the official fan should still be able to rotate

PR #60, opened 2023-06-26
iwlwifi: pnvm: pretend to have loaded pnvm file on 5.10 bsp

nyanmisaka:
The pnvm file will reappear when updating `armbian-firmware-full` or `linux-firmware`.

"Fix" it in the kernel to save users the hassle of having to remove it manually.

This should be a better solution for Intel AX210 users, before the 6.1 LTS kernel comes.

### DIFF
--- a/drivers/net/wireless/intel/iwlwifi/fw/pnvm.c
+++ b/drivers/net/wireless/intel/iwlwifi/fw/pnvm.c
@@ -242,6 +242,14 @@ int iwl_pnvm_load(struct iwl_trans *trans,
 
 	/* load from disk only if we haven't done it (or tried) before */
 	if (!trans->pnvm_loaded) {
+#if 1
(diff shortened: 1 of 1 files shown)


amazingfate:
So pnvm firmware will never get loaded. Does this has side effect on other wireless cards?

amazingfate:
AX211 and AX411 also need to load pnvm firmware, do they have the same issue as AX210?
Maybe we can only bypass pnvm_load for AX210 by `trans->cfg->name == "Intel(R) Wi-Fi 6 AX210 160MHz"`.

nyanmisaka:
AX211 and AX411 are only supported on intel x86 processors via CnVi interface. They are not available on Rockchip platforms.

amazingfate (approved):

nyanmisaka:
I have no idea about the failed CI. It builds fine locally.

amazingfate:
Nevermind that fails because of lacking upload permission, kernel build is successful.

PR #63, opened 2023-07-04
arm64: dts: rock5a: providing more cooling levels for pwm-fan

amazingfate:
Almost the same @FantasyGmm has done in #43 .
It works without modifying `&threshold`.
`sustainable-power` is a property for policy power_allocator so is should be no harm here, no need to delete.


igorpecovnik (approved):
lgtm

PR #66, opened 2023-07-06
workflow: don't upload kernel in pr

amazingfate:


amazingfate:
I find that `ARTIFACT_USE_CACHE=yes` will build kernel without uploading to ghcr. We can download build results from the workflow artifacts like https://github.com/armbian/linux-rockchip/suites/14104061407/artifacts/789044188

igorpecovnik (approved):
good catch, didn't notice - if it works, lets merge

PR #71, opened 2023-07-20
fix indiedroid wifi DTS by removing WIFI,poweren_gpio

lanefu:
9tripod device tree looking for a wifi switch, when we don't use on indiedroid... this device tree is only used for indiedroid.. removed line


lanefu:
pairs nicely with https://github.com/armbian/build/pull/5493

igorpecovnik (approved):

efectn (approved):

PR #72, opened 2023-07-21
Fix NanoPi R6 USB3 Issue

efectn:
check:
- https://github.com/Joshua-Riek/ubuntu-rockchip/issues/258
- https://forum.armbian.com/topic/29273-regression-usb-30-port-not-working-with-kernel-510160/


amazingfate:
The reverted upstream commit does fix one issue, does this revert have other effects?

efectn:
> The reverted upstream commit does fix one issue, does this revert have other effects?

Agree will test it. Normally it shouldn't cause any problem due to it wasn't exists on rkr3. Also try to find better fix instead of this hack

efectn:
@amazingfate done

amazingfate (approved):

PR #73, opened 2023-07-22
arm64: dts: blade3: add sata2 overlay for mini pcie port

Joshua-Riek:
The Mixtile Blade 3 has a Mini PCIe port, and its position on the board makes it perfect for an mSATA SSD. So here is an overlay to enable SATA2 instead of PCIe.


amazingfate (approved):

rpardini (approved):

rpardini:
Hey @Joshua-Riek good stuff. I was considering the other day what to put in that slot, not so many mini pcie cards available these days.
Do you think a Kingston KC600 or such would work?

Joshua-Riek:
I actually tested this out with a 256GB Kingston KC600 and it works perfectly!

PR #76, opened 2023-07-29
arm64: dts: opi5: make type-c dr_mode host as default

efectn:


igorpecovnik (approved):

arch/arm64/boot/dts/rockchip/rk3588s-orangepi-5.dtsi
```diff
@@ -361,7 +361,7 @@
 
 &usbdrd_dwc3_0 {
 	status = "okay";
-	dr_mode = "otg";
+	dr_mode = "host";
 	usb-role-switch;
```
hzyitc, line 365: `usb-role-switch` only work with `dr_mode = "otg";` ( Not sure )

https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/usb/usb-drd.yaml#L51-L63

You can use the following config:

```
&usbdrd_dwc3_0 {
	status = "okay";
	dr_mode = "otg";
	usb-role-switch;
	role-switch-default-mode = "host";
};
```

hzyitc:
A good example for otg is:

Some boards might use USB-A female connector for USB ports, however,
the port could be connected to a dual-mode USB controller, making it
also behaves as a peripheral device if male-to-male cable is connected.

arch/arm64/boot/dts/rockchip/rk3588s-orangepi-5.dtsi
efectn: Are there any way to produce it works/not works?

hzyitc: Just try to switch mode:

Not sure if the command is correct. My device have some problems now.

```
echo device >/sys/class/usb_role/usbxxx/role
```

What's more, you can configure the gadget and test with a male-to-male cable.

efectn: You seem right i can't assign device role.

I've also tried your way but host didn't work

efectn: I think overlay is fine for who wants to use OTG mode

arch/arm64/boot/dts/rockchip/rk3588s-orangepi.dtsi
Joshua-Riek, line 1: The Orange Pi 5 Plus does not need the USB2 init service, so it would be best to leave this as `otg`. Unless this fixes another issue I'm not aware of.

arch/arm64/boot/dts/rockchip/rk3588s-orangepi-5.dtsi
hzyitc: > You seem right i can't assign device role.
> 
> I've also tried your way but host didn't work

What's it mean?

Can't switch or fail to connect to the host device.

I'm not sure if `device` is corrent. I will test it.

> I think overlay is fine for who wants to use OTG mode

It's depend on the maintainers. I just give some suggests.

efectn: @hzyitc `/sys/class/usb_role/fc000000/role` is not exists. Howeve, `/sys/kernel/debug/usb/fc000000.usb/mode` is exists and it prints `device` if i check it with cat despite i add `role-switch-default-mode = "host";`

amazingfate:
Close as we are using rkr6 branch now.

PR #79, opened 2023-08-14
arm64: dts: nanopct6: remove fspim1 pin definitions

Joshua-Riek:
Hello, I have a fix for my previous pr #78, as `/dev/mtdblock0` would still not appear. I read the schematics wrong, and the fspim1 pin definitions must be removed.

I finally got a T6 with working SPI and eMMC, so I can confirm this patch will allow for `/dev/mtdblock0` to show as expected.


amazingfate:
Looking good to me. CI failed because of low disk space.

amazingfate (approved):

PR #86, opened 2023-08-26
fix(git): ignore some compiled products generated by the kernel

sora1874:
git status will show：
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,16 @@
 *.tar
 *.xz
 *.zst
+*.cmd
+*.d
+*.S
+*.config
+*.config.old
+*.tmp_System.map
+*.tmp_kallsyms
+*.tmp_vmlinux.kallsyms1
+*.tmp_vmlinux.kallsyms2
+*.version
 Module.symvers
 modules.builtin
 modules.order


.gitignore
```diff
@@ -47,6 +47,16 @@
 *.tar
 *.xz
 *.zst
+*.cmd
+*.d
+*.S
+*.config
```
amazingfate, line 53: There are files tracked by git.

amazingfate (changes requested):

.gitignore
amazingfate, line 53: there are files `*d`, `*.S` and `*.config` tracked by git.

sora1874: image
is not use .gitignore file ?

amazingfate: test.d is a dir: https://github.com/armbian/linux-rockchip/tree/rk-5.10-rkr5.1/tools/testing/selftests/ftrace/test.d
There are already *.S in git: https://github.com/armbian/linux-rockchip/blob/rk-5.10-rkr5.1/usr/initramfs_data.S
Also *.config: https://github.com/armbian/linux-rockchip/blob/rk-5.10-rkr5.1/tools/testing/selftests/wireguard/qemu/kernel.config

PR #88, opened 2023-08-30
arm64: dts: add devicetree of orangepi3b

amazingfate:
Devicetree copied from https://github.com/orangepi-xunlong/linux-orangepi/blob/orange-pi-5.10-rk35xx/arch/arm64/boot/dts/rockchip/rk3566-orangepi-3b.dts
Add panfrost driver support: https://github.com/amazingfate/linux-rockchip/blob/opi3b/arch/arm64/boot/dts/rockchip/rk3566-orangepi-3b.dts#L341-L342


amazingfate:
@igorpecovnik Our workflow has a lot of kernel source cache: https://github.com/armbian/linux-rockchip/actions/runs/6021065148/job/16333313290#step:4:29, which are causing workflow fail to run.
In this PR I delete some big size and ununsed source for more place.

igorpecovnik:
Oh, it was at 70% ... strange that it failed. Well, cleaning of runners has to be improved I guess. Deleting everything is easy ... but then caching doesn't make sense. Cleaned now.

amazingfate:
Tested on opi3b, everything should be the same with orangepi's kernel except this change:
I removed `cursor-win-id = <ROCKCHIP_VOP2_CLUSTER0>;` from devicetree node `&vp0`. This property will make mouse cursor invisible under wayland.

rpardini (approved):

PR #95, opened 2023-09-27
arm64: dts: orangepi3b: remove edp-panel & backlight

rpardini:
- those are just deferred forever; whoever needs this might do an overlay


rpardini:
Nah, my brain's not working right.

PR #101, opened 2023-10-15
arm64: dts: turing-rk1: device tree cleanup and use correct stdout-path

Joshua-Riek:
I cleaned up the device tree for the Turing RK1 and adjusted the stdout path to use UART 9.

This device uses UART 9 with a baud rate of 115200 for the console instead of the standard UART 2 with a baud rate 1500000 for Rockchip devices. I wouldn't say I like this, but it is necessary for the hardware..


amazingfate (approved):

rpardini:
Hey @Joshua-Riek, what if in the `aliases { ... }` you'd invert uart2 and uart9, so that `/dev/ttyS2` would be kept as debug UART? It might reduce the surprise factor for users...

Joshua-Riek:
I was able to get UART2 and UART9 working at the same time, I made a weird mistake with U-Boot that was causing the problem. However, I would like to stay close to the proposed mainline patch for this board. I have already expressed my concerns to the author, and hopefully, it will be considered.

https://lore.kernel.org/lkml/20231007043600.151908-4-CFSworks@gmail.com/T/

PR #104, opened 2023-10-20
video: rockchip: rga3: fix compatibility with legacy global alpha config

nyanmisaka:
Fixes a RGA regression in **rkr6** 7291693, which broke compatibility with **librga 1.9.x and older** when using DST over blend/composite and premultiplied alpha.

It's a typo.
```c
req->alpha_config.bg_global_alpha_value = 0xff; // should be `fg` foreground instead of `bg` background
req->alpha_config.bg_global_alpha_value = 0xff; // duplicated
```

When `req_rga->feature.global_alpha_en` is false, it means the API version of userspace librga is 1.9.x and older. And this fix is to make sure it falls back to the legacy behavior.

In addition, the source code of librga 1.10 is not yet available to the public at this point.

### DIFF
--- a/drivers/video/rockchip/rga3/rga2_reg_info.c
+++ b/drivers/video/rockchip/rga3/rga2_reg_info.c
@@ -2108,7 +2108,7 @@ static void rga_cmd_to_rga2_cmd(struct rga_scheduler_t *scheduler,
 					req->alpha_config.bg_global_alpha_value = 0xff;
 				}
 			} else {
-				req->alpha_config.bg_global_alpha_value = 0xff;
+				req->alpha_config.fg_global_alpha_value = 0xff;
 				req->alpha_config.bg_global_alpha_value = 0xff;
 			}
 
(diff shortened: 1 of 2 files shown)


ginkage (approved):

amazingfate (approved):

amazingfate:
Another rga3 related issue: gstreamer videoconvert can't use RGA3 to convert hdmiin. Here is the pipeline command line I use:
```
GST_VIDEO_CONVERT_USE_RGA=1 gst-launch-1.0 -e v4l2src device=/dev/video0 ! videoconvert ! waylandsink
```
After reverting https://github.com/armbian/linux-rockchip/commit/fb4fd88e2200b9b8d135f9ad4b984ea2b5047556 this pipeline will work again. But I haven't found the root cause.

nyanmisaka:
> Another rga3 related issue: gstreamer videoconvert can't use RGA3 to convert hdmiin. Here is the pipeline command line I use:
> 
> ```
> GST_VIDEO_CONVERT_USE_RGA=1 gst-launch-1.0 -e v4l2src device=/dev/video0 ! videoconvert ! waylandsink
> ```
> 
> After reverting [fb4fd88](https://github.com/armbian/linux-rockchip/commit/fb4fd88e2200b9b8d135f9ad4b984ea2b5047556) this pipeline will work again. But I haven't found the root cause.

It seems to be related to the use of `virt_addr`, I don't use it in my FFmpeg build and use `dma_fd` instead, so I don't encounter this problem. Maybe the maintainer of gstreamer plugin @JeffyCN knows more.

PR #109, opened 2023-10-24
arm64: dts: turing-rk1: remove usb controller and phy nodes for usb c

Joshua-Riek:
These nodes are required for USB C, which the Turing RK1 does not have. So they should be removed.

Thanks!


Joshua-Riek:
Ignore this pr. The pins on the SoM used for USB C are connected to the USB 2.0 OTG port.

PR #111, opened 2023-10-24
drivers: bluetooth: add support for RTL8852BE

Joshua-Riek:
This PR fixes Bluetooth for the RTL8852BE, tested with the Orange Pi 5 Plus WiFi + Bluetooth module.

Cherry-picked from: https://github.com/orangepi-xunlong/linux-orangepi/commit/7e6c3163aa7e58b19730aa2aa259f1bb957cbca0
Firmware PR: https://github.com/armbian/firmware/pull/67

### DIFF
--- a/drivers/bluetooth/btrtl.c
+++ b/drivers/bluetooth/btrtl.c
@@ -25,6 +25,7 @@
 #define RTL_ROM_LMP_8821A	0x8821
 #define RTL_ROM_LMP_8761A	0x8761
 #define RTL_ROM_LMP_8822B	0x8822
+#define RTL_ROM_LMP_8852A	0x8852
 #define RTL_CONFIG_MAGIC	0x8723ab55
 
 #define IC_MATCH_FL_LMPSUBV	(1 << 0)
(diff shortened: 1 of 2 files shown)


amazingfate (dismissed):

amazingfate:
mainline's commit to support RTL8852B: https://github.com/torvalds/linux/commit/18e8055c88142d8f6e23ebdc38c126ec37844e5d
Is it possiable to use the firmware from mainline?

rpardini:
True. We've had similar during rtw88 migration on mainline, I think with 8822CS SDIO; the old driver matches with old firmware, while the mainline needs newer firmware.
In this case, a viable hack would be to rename prefix `rtl_bt/rtl8852bu` to `rtl_bt/rtl8852bu_vendor` (or some such) in both the kernel source and fw filenames?

Joshua-Riek:
I will test out the mainline firmware tonight and see if it can work. If the vendor firmware is required, I think adding a vendor prefix will be a good alternative.

Joshua-Riek:
I just confirmed we can use the firmware from mainline. So, this should be good to merge. I will close  https://github.com/armbian/firmware/pull/67.

Thanks!

Joshua-Riek:
I think that I misunderstood. Should I send a PR to the firmware repo with the latest mainline RTL8852BE firmware?

amazingfate:
No need to add mainline firmware to armbian/firmware, armbian-firmware-full will package the latest mainline firmware.

amazingfate (approved):

PR #123, opened 2023-11-30
arm64: dts: Add support for roc-rk3588s-pc

zouxf:
roc_rk3588s_pc is a RK3588S-based board, the vendor is firefly, the mini-pc station-m3 is based on this board.


amazingfate:
`rk3588s-roc-rk3588s-pc-v12.dts` and `rk3588s-roc-rk3588s-pc-v12.dtsi` are sharing the same name, so why split them into two files?

zouxf:
Since there is also a v13 for the roc-rk3588s-pc, in the future I would like the dts for v13 to include the v12.dtsi and the differences from v12. It sounds like using v12.dtsi as a base dtsi and the dts file including this file and the differences to v12.dtsi.

amazingfate:
What about using a singe `rk3588s-roc-rk3588s-pc-v12.dts` for v12 board, and include it in the future v13?

zouxf:
Okay, I'll fix this. Thank you!

amazingfate (approved):

efectn (approved):

PR #127, opened 2023-12-06
Add Hinlink HT2 dts and aic8800 wifi driver

hoochiwetech:
Hinlink ht2 is based on Hinlink h28k.
deletes 1 gigabit port and adds 1 HDMI output
1 HDMI3.0 host and 1 wifi module


amazingfate:
Please split wireless driver and devicetree change into two commits. And there is no need to change rockchip_linux_defconfig.

PR #139, opened 2024-01-08
media: rockchip: hdmirx: add VIDIOC_G_PARM

amazingfate:
This will let obs linux-v4l2 plugin show display from hdmirx of rk3588.
OBS code using these API: https://github.com/obsproject/obs-studio/blob/master/plugins/linux-v4l2/v4l2-helpers.c#L244-L254

@ramiropolla @rigaya

### DIFF
--- a/drivers/media/platform/rockchip/hdmirx/rk_hdmirx.c
+++ b/drivers/media/platform/rockchip/hdmirx/rk_hdmirx.c
@@ -1625,6 +1625,23 @@ static int hdmirx_s_input(struct file *file, void *priv, unsigned int i)
 	return i == 0 ? 0 : -EINVAL;
 }
 
+static int hdmirx_g_parm(struct file *file, void *priv,
+		struct v4l2_streamparm *parm)
+{
+	struct hdmirx_stream *stream = video_drvdata(file);
+	struct rk_hdmirx_dev *hdmirx_dev = stream->hdmirx_dev;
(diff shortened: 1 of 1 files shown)


ramiropolla (changes requested):
I'm not a kernel dev, but after a quick glance and checking how other drivers do it, I see at least two issues:
- hdmirx_g_parm() is not checking for parm->type like other drivers
- if changing parameters is not supported, then we shouldn't implement vidioc_s_parm() (although I see there's a function called hdmirx_s_dv_timings())
- I'm not sure about the readbuffers. where did you get that info?

amazingfate:
I don't know readbuffers value exactly, it just works. Maybe not returning it will work, I will test.
I will add check for type `V4L2_BUF_TYPE_VIDEO_CAPTURE` and remove vidioc_s_parm to see if it will work.

amazingfate:
Add checking `parm->type` is `V4L2_BUF_TYPE_VIDEO_CAPTURE_MPLANE`, remove `vidioc_s_parm` and return of `readbuffers`.
Tested on obs with mesa llvmpipe driver, panfork driver failed with gl call error.

amazingfate:
This commit is not that useful on 5.10 kernel because GPU support is bad. Since we have panthor support on 6.1, I will test with panthor and create a pr for 6.1 kernel.

PR #142, opened 2024-01-21
fs: Remove weird export restrictions

simonswine:
This prevents external file system (like ZFS for me) from being built, roughly automated creation of the changeset like this:

```
$ git ls-files | xargs ag -l ANDROID_GKI_VFS > files
$ cat files | xargs sed -i '/MODULE_IMPORT_NS(ANDROID_GKI_VFS_EXPORT_ONLY);/d' 
$ cat files | xargs sed -i 's/EXPORT_SYMBOL_NS(\(.*\), ANDROID_GKI_VFS_EXPORT_ONLY)/EXPORT_SYMBOL(\1)/g'
```


amazingfate:
These are from an android commit: https://github.com/armbian/linux-rockchip/commit/d483eed85ffb9879b103cd8d6b3fae367fa6193e. What about reverting it?

simonswine:
> These are from an android commit: [d483eed](https://github.com/armbian/linux-rockchip/commit/d483eed85ffb9879b103cd8d6b3fae367fa6193e). What about reverting it?

I imagined it would be harder to revert this, but I had not tried. Now used revert and a couple of manual clean-ups. Take another look please.

amazingfate:
Commit `https://github.com/armbian/linux-rockchip/commit/d483eed85ffb9879b103cd8d6b3fae367fa6193e` has modified 96 files, but your revert modifies 124 files, where does other files come from？

Joshua-Riek:
> I imagined it would be harder to revert this, but I had not tried. Now used revert and a couple of manual clean-ups. Take another look please.

I think they squashed a few changes into the revert commit. If this is the case I would create two commits, one to revert https://github.com/armbian/linux-rockchip/commit/d483eed85ffb9879b103cd8d6b3fae367fa6193e, then another for any other changes needed for ZFS to compile.

amazingfate:
First revert https://github.com/armbian/linux-rockchip/commit/eff1ffbf0c097512fe2c92462b6183f716afb770, then https://github.com/armbian/linux-rockchip/commit/d483eed85ffb9879b103cd8d6b3fae367fa6193e. I don't know if this will work.

amazingfate:
Close as lacking activity.

hufman:
Thank you for this reference, it fixes my local kernel build

PR #145, opened 2024-01-27
arm64: dts: rock-5a: pull down data-strobe to fix emmc compatibility

nyanmisaka:
[Mainline has enabled this](https://patchwork.kernel.org/project/linux-rockchip/patch/20231205202900.4617-2-CFSworks@gmail.com/) globally for all 3588 boards. I'm not sure if this is a good idea for the BSP kernel as well.
So only enable it on Rock-5a which is known to be affected.

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3588s-rock-5a.dts
+++ b/arch/arm64/boot/dts/rockchip/rk3588s-rock-5a.dts
@@ -892,6 +892,12 @@
 			rockchip,pins = <0 RK_PC4 RK_FUNC_GPIO &pcfg_pull_none>;
 		};
 	};
+
+	emmc {
+		emmc_data_strobe: emmc-data-strobe {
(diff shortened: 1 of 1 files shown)


amazingfate (approved):

Joshua-Riek:
I know the Turing RK1 has this problem, it's how the mainline patch was found. If more boards have eMMC troubles in the future, it may be a good idea to enable the patch globally.

nyanmisaka:
> As a result, many RK3588 board designers do not bother adding a dedicated RDS resistor, instead relying on the RK3588's internal bias.

Affected boards appear to have certain resistors omitted from the circuit design. The issue will arise when the onboard eMMC or user-supplied eMMC module is from a specific vendor. Samsung and Kioxia are known to be affected, while FORESEE and SanDisk are very lucky.

PR #147, opened 2024-01-30
Update armsom-sige7 device tree

Z-Keven:
Update armsom-sige7 device tree to solve some problems


Z-Keven:
I'll close the submission first because there are still things that need to be corrected

Issue #96, opened 2023-10-02
Hinlink H66K/H68K - missing GPIO LEDs in DTS.


ddscentral:
Hinlink H66K/H68K boards have 3 GPIO LEDs:
* Pin RK_PB0 - green
* Pin RK_PA5 - blue
* Pin RK_PA7 - amber 

However, in the rk3568-hinlink-h6xk.dtsi, only the green LED is defined (as "work").
Manually adding the other two to the dts file and recompiling it makes all 3 LEDs work correctly.

PR #152, opened 2024-02-13
arm64: dts: add missing led gpio

amazingfate:
#96
Not yet tested, ref from https://github.com/unifreq/linux-6.1.y/blob/main/arch/arm64/boot/dts/rockchip/rk3568-hlink-h68k.dts#L57-L76


amazingfate:
Tested on hinlink h66k, 3 leds work.

PR #157, opened 2024-03-15
driver: rknpu: Fixing deadlock issue with spin_lock in interrupt hand…

amazingfate:
There is a npu deadlock bug in rkr6 kernel, fixed in rkr7.1. 6.1-rkr1 is not affected.


Joshua-Riek:
Looks good, the NPU no longer crashes for me with this patch.

PR #171, opened 2024-04-21
Add mainline panthor driver to 6.1 rkr1

amazingfate:
I squashed commits from https://github.com/rpardini/armbian-linux-rockchip-rk3588/tree/armbian-rk-6.1-rkr1-plus-boogie-panthor-v6, delete 3 dts commits and keep arm mali gpu driver as default.
To enable panthor driver, use `armbian-config` to enable `panthor-gpu`.


amazingfate:
This pr has updated driver code of lima and panfrost, so test on rk356x and rk3528 is necessary.

amazingfate:
Tested with hinlink ht2(rk3528) and rock3a(rk3568), this should not harm lima and panfrost driver.

nyanmisaka:
> and keep arm mali gpu driver as default.

Note that OpenCL/libmali and panfork still require bifrost. So making panthor opt-in makes sense.

PR #175, opened 2024-04-24
Add dynamic-power-coefficient properties to all cores

ThomasKaiser:
Correct dynamic-power-coefficient properties for the cores are a prerequisit for schedutil/EAS to work correctly.

RK's 100/300 values are BS of course, just compare with RK3399: https://patchwork.kernel.org/project/linux-arm-kernel/patch/1500974575-2244-1-git-send-email-wxt@rock-chips.com/#20763985

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3588s.dtsi
+++ b/arch/arm64/boot/dts/rockchip/rk3588s.dtsi
@@ -452,6 +452,7 @@
 			d-cache-line-size = <64>;
 			d-cache-sets = <128>;
 			next-level-cache = <&l2_cache_l1>;
+			dynamic-power-coefficient = <100>;
 		};
 
 		cpu_l2: cpu@200 {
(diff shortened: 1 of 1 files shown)


amazingfate (approved):

ThomasKaiser:
The `capacity-dmips-mhz` values [are BS too](https://github.com/armbian/build/pull/6120#issuecomment-2075003397). If this DMIPS nonsense should be followed then according to [my own measurements](https://github.com/ThomasKaiser/sbc-bench/blob/master/Benchmarking_some_benchmarks.md#dhrystone--dmips--dmipsmhz) dhrystone attests the A76 being not 1.9 times but 2.32 times faster than the A55 in RK3588[s], as such it should read for the A55:

    capacity-dmips-mhz = <441>;

But ofc this is stupid since everybody (should) know that DMIPS from last century are something that doesn't represent anything we do today on these machines. And the whole idea of EAS on systems mixing in-order and out-of-order cores is flawed by design :)

But for `schedutil` to have at least a chance to somehow work correctly with EAS in mind some properties have to be set as best as possible (and RK's BSP defaults are just weird)

PR #186, opened 2024-05-14
add rk3566-firefly-roc-pc 5.10 dts

chainsx:


.config
amazingfate, line 1: It's better to delete this file.

amazingfate (changes requested):

.config
chainsx: I think the ".config" file should be added to the ".gitignore" file?

efectn (approved):

PR #207, opened 2024-07-14
6.1-rkr1: sync FriendlyElec's DTs from vendor (common, R6S, R6C, T6, CM3588) + add T6-LTS DT

rpardini:
#### arm64: dts: sync FriendlyElec's DTs from vendor source (common, R6S, R6C, T6, CM3588)

- From vendor's tree at https://github.com/friendlyarm/kernel-rockchip branch `nanopi6-v6.1.y` sha1 `925eb8ba4c765d4282a7281ddc6559a6e45c2687`
- adapt nomenclature, pull changes that seem to make sense, drop the ones we've explicitly done in Armbian fork

#### arm64: dts: add FriendlyElec's `rk3588-nanopc-t6-lts` DT

- the "T6 LTS" has a different regulator for USB2
- From vendor's tree at https://github.com/friendlyarm/kernel-rockchip branch `nanopi6-v6.1.y` sha1 `925eb8ba4c765d4282a7281ddc6559a6e45c2687`
- Adapted vendor's nomenclature (rev07 -> rk3588-nanopc-t6-lts.dts)
  - adapted include (rev01 -> rk3588-nanopc-t6.dts)
  - dropped `&mach` node that is vendor-specific
- this requires the `vcc5v0_host20_en` node added in previous commit to the `t6` DT
- add to Makefile

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3588s-nanopi-r6-common.dtsi
+++ b/arch/arm64/boot/dts/rockchip/rk3588s-nanopi-r6-common.dtsi
@@ -116,6 +116,10 @@
 	status = "okay";
 };
 
+&avsd {
+	status = "okay";
+};
+
 &combphy0_ps {
 	status = "okay";
 };
@@ -184,9 +188,9 @@
 };
 
 &hdmi0 {
+	cec-enable;
 	enable-gpios = <&gpio4 RK_PB6 GPIO_ACTIVE_HIGH>;
 	status = "okay";
-	cec-enable = "true";
 };
 
 &hdmi0_in_vp0 {
@@ -526,6 +530,10 @@
 	status = "okay";
 };
 
+&rkvtunnel {
+	status = "okay";
+};
+
 &avcc_1v8_s0 {
 	regulator-state-mem {
 		regulator-on-in-suspend;
@@ -667,6 +675,8 @@
 &vop {
 	assigned-clocks = <&cru ACLK_VOP>;
 	assigned-clock-rates = <800000000>;
+	vop-supply = <&vdd_log_s0>;
+	support-multi-area;
 	status = "okay";
 	disable-win-move;
 };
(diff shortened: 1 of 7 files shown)


arch/arm64/boot/dts/rockchip/rk3588s-nanopi-r6-common.dtsi
```diff
@@ -708,3 +741,10 @@
 	clocks = <&hdptxphy_hdmi_clk0>;
 	clock-names = "hdmi0_phy_pll";
 };
+
+/* @TODO: FE has display_subsystem different:
+&display_subsystem {
+	clocks = <&hdptxphy_hdmi0>, <&hdptxphy_hdmi0>;
+	clock-names = "hdmi0_phy_pll", "hdmi1_phy_pll";
```
efectn, line 748: I think this is only necessary for the boards have 2 HDMI outs

rpardini: Makes sense. I guess FriendlyElec devs are economizing: more devices have 2 HDMI inputs so they rather duplicate/override those that _don't_.

rpardini:
built and tested on-device (T6 LTS), where this does fix the USB2. (soon to go figure out mainline equiv, but no usb (2/3/C) works there at all so I wanted to check with vendor kernel)

I could distill this to the new DT for the LTS version and move the pin definition (that FE did at the base T6) there and keep this isolated if ya'll prefer.

I'm unsure and can't really test the other changes/boards, which seem fairly minor: edid, fans, enable some 40-pin pins, trcm-sync-tx-only (?) and global `avsd` and `rkvtunnel` -- no idea what those are.

let me know.

nyanmisaka:
> and global avsd and rkvtunnel -- no idea what those are.

avs2 video decoder => a [Chinese video codec](https://en.wikipedia.org/wiki/Audio_Video_Standard) to avoid patent issues in h26x codec, performance similar to hevc/h265.

If this is not enabled, in recent [mpp](https://github.com/rockchip-linux/mpp) libraries, a warning will be thrown in journalctl even when not using avs2. So ideally it would be best to enable it for all rk3588 boards.
```
rock-5a mpp[3336]: mpp_platform: client 12 driver is not ready!
```

rkvtunnel => [ROCKCHIP videotunnel driver](https://github.com/armbian/linux-rockchip/blob/rk-6.1-rkr1/drivers/video/rockchip/vtunnel/rkvtunnel.c), not sure how to use too

amazingfate:
We have bumped to 6.1 rkr3, closing now.

PR #220, opened 2024-08-13
arm64: dts: rockchip: Adopt mainline naming for Friendlyelec CM3588 NAS

ColorfulRhino:
Rename the dts file to match the new name from mainline kernel 6.11.
This makes it easier for the board config since no conditionals are needed.

~~Should be merged at the same time with https://github.com/armbian/build/pull/7082~~ (Due to the added compatibility patch, merging at the same time is not needed anymore)

This only applies to branch rk-6.1-rkr3 on purpose since old versions are deprecated.


amazingfate (approved):

arch/arm64/boot/dts/rockchip/Makefile
```diff
@@ -343,7 +343,7 @@ dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3588-armsom-w3.dtb
 dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3588-firefly-itx-3588j.dtb
 dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3588-fxblox-rk1.dtb
 dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3588-hinlink-h88k.dtb
-dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3588-nanopc-cm3588-nas.dtb
```
efectn, line 352: I don't think this is a backward-compatibility change. What about adding fallback devicetree for the old naming?

ColorfulRhino: @efectn  Backwards compatible in which way do you mean?

efectn: > @efectn Backwards compatible in which way do you mean?

Systems using this devicetree name will not boot anymore after updating kernel

ColorfulRhino: I see the problem now, thanks. This can probably be solved with a symlink of some sort.

ColorfulRhino: I have added a link/stub dts with the old name. This _should_ work for existing installations.

amazingfate (approved):

amazingfate:
LGTM, feel free to merge when https://github.com/armbian/build/pull/7082 is done.

efectn (approved):

ColorfulRhino:
Let's merge this now since due to the added compatibility support, both dts names are working and thus merging at the same time with the above mentioned PR is not needed anymore.
@amazingfate You need to merge since you have write access :) I have rebased it onto the current branch.

amazingfate (approved):

PR #243, opened 2024-09-18
Added device tree overlays files for armsom-sige7 board:two OV13850 c…

itlhd:
…ameras

Two new cameras have been adapted on Sige7, and I would like to submit them here.

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3528-armsom-sige1.dts
+++ b/arch/arm64/boot/dts/rockchip/rk3528-armsom-sige1.dts
@@ -253,7 +253,7 @@
 	sdio_pwrseq: sdio-pwrseq {
 		compatible = "mmc-pwrseq-simple";
 		pinctrl-names = "default";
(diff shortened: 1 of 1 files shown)


arch/arm64/boot/dts/rockchip/overlay/Makefile
```diff
@@ -2,6 +2,8 @@
 dtbo-$(CONFIG_ARCH_ROCKCHIP) += \
 	armsom-sige7-camera-imx415-4k.dtbo \
 	armsom-sige7-display-10hd.dtbo \
+	armsom-sige7-camera-ov13850-csi0.dtbo \
+	armsom-sige7-camera-ov13850-csi1.dtbo \
```
amazingfate: Please add these two dtbo after `armsom-sige7-camera-imx415-4k.dtbo`, so that Makefile is sorted by the order of alphabet.

amazingfate (changes requested):

PR #247, opened 2024-09-25
arm64: dts: Add missing overlays for armsom-sige3

itlhd:
Add missing overlays for armsom-sige3. 
These configurations are universal, I think others should also be able to use them.


amazingfate:
I will test `armsom-sige3-display-10hd.dtbo` and `armsom-sige3-camera-ov13850.dtbo`, which are sige3 only overlays.

PR #254, opened 2024-10-02
arm64: dts: DTS for Luckfox Core3566

buldo:
Device tree files from vendor SDK
Link to SDK https://drive.google.com/drive/folders/1q8zXpt3WzSfGyjJrTL8ObEmExVmXp-Ag

P.S. This files are aweful. But I have no enought knowledges to improve it.

P.P.S. Tested. Board works.
Related issue https://github.com/armbian/build/issues/7295


SuperKali (changes requested):
Hi, the CI fails as the file corresponding to the board dts is missing:

```bash
  make[3]: *** No rule to make target 'arch/arm64/boot/dts/rockchip/rk3566-luckfox-core3566.dtb', needed by 'arch/arm64/boot/dts/rockchip/'.  Stop.
   make[3]: *** Waiting for unfinished jobs....
```

I suggest you review the dts, usually the ones provided by the vendor are not put together so well and are quite confusing, which leads you to rewrite them from 0.

rpardini:
Yeah and at least rename (or use a placeholder include) to adhere to the naming convention that exists in the directory. Also maybe just gather all includes in a single thing.

buldo:
@SuperKali I forgot most important dts. Added in the second commit.  
@rpardini are there special command for dts preprocess? I mean to prepare single dts with all includes?

And one more question:  
`core3566-mipi-dsi*` files are related to dsi imege output. Vendor just provides 3 different sd card images. What should I do with this files?

SuperKali:
> `core3566-mipi-dsi*` files are related to dsi imege output.

I think they cannot be activated at the same time and one of them must be disabled

> Vendor just provides 3 different sd card images. What should I do with this files?

I usually recommend creating a single file and integrating an if function, although it would make sense to create an overlay for the latter

buldo:
@SuperKali will try to provide single dts after this weekend

buldo:
@SuperKali now dts is single file.  
I tested it with my board with HDMI and Ethernet.  
My board have no emmc and WiFi/BT so I not able to test it.

SuperKali (approved):
Ok, perfect!

PR #257, opened 2024-10-04
add board mekotronics r57 rk3576

monkaBlyat:


amazingfate:
It's better to add changes in Makefile and devicetree into one commit.

rpardini:
We _could_ squash commits when merging, though.

amazingfate (approved):

PR #262, opened 2024-10-11
add itc type to hdmirx

benhoff:
This backports the ability to check the ITC type.

### DIFF
--- a/drivers/media/platform/rockchip/hdmirx/rk_hdmirx.c
+++ b/drivers/media/platform/rockchip/hdmirx/rk_hdmirx.c
@@ -226,6 +226,7 @@ struct rk_hdmirx_dev {
 	struct v4l2_ctrl *audio_present_ctrl;
 	struct v4l2_dv_timings timings;
 	struct gpio_desc *hdmirx_det_gpio;
+	struct v4l2_ctrl *content_type;
 	struct work_struct work_wdt_config;
 	struct delayed_work delayed_work_hotplug;
 	struct delayed_work delayed_work_res_change;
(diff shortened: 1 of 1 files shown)


amazingfate (approved):

amazingfate:
I just tested your patch on rock5b and 6.11 mainline kernel, obs is not working. After removing your patch obs is fine.
I guess this would also break vendor kernel.

benhoff:
> I just tested your patch on rock5b and 6.11 mainline kernel, obs is not working. After removing your patch obs is fine. I guess this would also break vendor kernel.

@amazingfate  Thanks for testing!

It's running on the vendor kernel for me, but with my own custom v4l2 application, I hadn't tested obs as I'm not running a desktop.

 Can you help me understand the environment you are testing it in so I can recreate?

 I'm surprised you can run drivers from 6.1 on 6.11, but I don't do a lot of kennel development so that's why I'm asking.

I'll spin up an environment and get to the bottom of obs here in a few days

amazingfate:
I was testing mainline kernel with [your commit](https://github.com/armbian/build/pull/7224) yesterday. 
I get `error: v4l2-input: /dev/video0: select timed out` when adding v4l2 device to obs. I'm running armbian noble.
Debian has broken obs because they add this patch: https://salsa.debian.org/multimedia-team/obs-studio/-/commit/4a353a071febf315746a9d184bc6506e31421882, I'm testing obs without this patch.
I tested 6.1 kernel with your patch today and find that there is still `select timed out` issue without your patch, so your patch should be not related. I will check mainline kernel tomorrow.

amazingfate:
This commit is fine, let's merge it.

PR #276, opened 2024-11-02
arch: arm64: dts: add new board armsom-cm5-cm4-io

amazingfate:
This is armsom cm5 `SoM` on raspberry cm4 carrier board.


amazingfate:
close since #281 has included it with newer commits

PR #285, opened 2024-11-11
fix : armsom-w3 screen and camera loading method

as-jackson:
armsom-w3 supports overlay, modify dts to overlay format


amazingfate:
Missing change in `overlay/Makefile`, dtbo is not built.

amazingfate (approved):

PR #290, opened 2024-11-18
driver: rknpu: Update rknpu driver, version: 0.9.8

### LINKED ISSUE #266: Add the new RKNPU kernel version 0.9.8
### Feature Description

As per the title, please add the new RKNPU kernel version 0.9.8 contained here https://github.com/airockchip/rknn-llm/tree/main/rknpu-driver

In the documentation, rockchip says that if the build fails modify the kernel header file kernel/include/linux/mm.h by adding the following code: 
```

static inline void vm_flags_set(struct vm_area_struct *vma, vm_flags_t flags){ 
vma->vm_flags |= flags; 
} 
static inline void vm_flags_clear(struct vm_area_struct *vma, vm_flags_t flags){ 
vma->vm_flags &= ~flags; 
} 
```

please update the releases with such kernel driver ASAP. If possible compile it as external module to be loaded with modprobe so if they update the driver, we can recompile it and load it with modprobe instead of recompiling the kernel.

Thank you

shirok1:
* Fix muti-process run error
* Fix mult-process domain switch error

Signed-off-by: Shuangjie Lin <shuangjie.lin@rock-chips.com>
Change-Id: I3052eca6ef577dc6420dd4c80a3ac7d0f17211ba

---

This was just a cherry-pick of https://github.com/armbian/linux-rockchip/commit/736d89f344156b75393b01ab2c0e3a06c39e110f and let's see if this works.

Closes #266.

### DIFF
--- a/drivers/rknpu/include/rknpu_drv.h
+++ b/drivers/rknpu/include/rknpu_drv.h
@@ -29,10 +29,10 @@
 
 #define DRIVER_NAME "rknpu"
 #define DRIVER_DESC "RKNPU driver"
-#define DRIVER_DATE "20240424"
+#define DRIVER_DATE "20240828"
 #define DRIVER_MAJOR 0
 #define DRIVER_MINOR 9
-#define DRIVER_PATCHLEVEL 7
+#define DRIVER_PATCHLEVEL 8
 
 #define LOG_TAG "RKNPU"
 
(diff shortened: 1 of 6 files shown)


shirok1:
Seems that `vm_flags_set` etc. was added in 6.3-rc1. If this patch do break the kernel we can actually consider patch the patch itself instead of adding a symbol to `mm.h` which is dirty.

amazingfate:
We are going to update to rkr4.1, which has rknpu 0.9.8.

shirok1:
Feel free to close this pr. I though [`rk-6.1-rkr3`](https://github.com/armbian/linux-rockchip/tree/rk-6.1-rkr3) was some "main" branch.

PR #300, opened 2024-12-06
drivers: mmc: add sdhci patch from rockchip

amazingfate:
There is a patch from rockchip to fix emmc issues.
EMMC of hinlink h66k/h68k/hnas is broken since rkr5.1, this will fix it.


igorpecovnik (approved):

amazingfate:
This patch can't fix h6xk emmc. I will investigate deeper.

PR #314, opened 2025-01-04
arm64: dts: rockchip: add SPI NOR flash overlay for NanoPi M6

efectn:


igorpecovnik:
i assume tested?

efectn:
> i assume tested?

Yes, tested

PR #320, opened 2025-01-24
Sync Panthor with drm-tip

ginkage:
Yet another round, this time bumping uAPI to 1.3.


EvilOlaf:
Would make sense to sync to rkr5, yes.

PR #323, opened 2025-01-28
net: ethernet: Fix motorcomm driver hardcoded "m" for CONFIG_FUXI

tizbac:


HeyMeco:
Can you tell how this fixes what and how this has been tested?

tizbac:
Without that patch if you are trying to get ethernet driver for example for orange pi 5 pro , to be built-in into kernel, even if you set y it will keep making ko file and so you will not have the driver in the kernel image...
It's standard code for all modules in the linux kernel, and rockchip probably always built it as module / it was an out of tree module before , so it was hardcoded "m" 

For armbian itself honestly does not make any difference because in standard config it is a module anyway, for me the issue surfaced because i tried to do what i said above :)

amazingfate (approved):

PR #339, opened 2025-02-21
station-m2: update bsp dts

chainsx:


chainsx:
https://github.com/armbian/build/pull/7858

PR #283, opened 2024-11-08
introcude devicetree property rockchip,default-link-up to fix rtl8125 on opi5plus

amazingfate:
Fixes #282 
First sync `rk_pcie_link_up` with mainline to fix some nvme devices not detected issue.
Second introcude devicetree property `rockchip,default-link-up` for opi5plus.
Tested on rock5b and with/without the property rtl8125 is always detected.

PR #340, opened 2025-03-03
Some commit for rock-2

CodeChenL:
The first commit: fixes https://github.com/armbian/build/issues/7836, test with rock-2a and rock-5c

### DIFF
--- a/drivers/pci/controller/dwc/pcie-dw-rockchip.c
+++ b/drivers/pci/controller/dwc/pcie-dw-rockchip.c
@@ -1475,7 +1475,10 @@ static int rk_pcie_hardware_io_config(struct rk_pcie *rk_pcie)
 			return ret;
 	}
 
-	reset_control_assert(rk_pcie->rsts);
+	if (device_property_read_bool(dev, "rockchip,skip-reset-in-config")) {
+		dev_info(dev, "skip reset controller\n");
+	} else
+		reset_control_assert(rk_pcie->rsts);
 	udelay(10);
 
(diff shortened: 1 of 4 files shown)


drivers/pci/controller/dwc/pcie-dw-rockchip.c
```diff
@@ -1475,9 +1475,6 @@ static int rk_pcie_hardware_io_config(struct rk_pcie *rk_pcie)
 			return ret;
 	}
 
-	reset_control_assert(rk_pcie->rsts);
-	udelay(10);
-
```
amazingfate, line 1480: This change will affect all rk socs. It should be device specific by introducing a devicetree property like this: https://github.com/armbian/linux-rockchip/pull/283

amazingfate:
>test with rock-2a and rock-5c

Does rock-5c has the same issue?
I will check if this will affect other rk3528 boards.

CodeChenL:
> > test with rock-2a and rock-5c
> 
> Does rock-5c has the same issue? I will check if this will affect other rk3528 boards.

The rock-5c doesn't have this problem, the purpose of this test item is to see if this commit will cause other soc's pcie to fail.

CodeChenL:
I've tested all rk3528 SBC from radxa with this problem (when pcie is turned on at the same time as gmac, the netowrk cord can't communicate, but the logs are all fine), it would be nice if you have other same devices to test with!

amazingfate:
I have checked that this will fix the same issue on armsom sige1 and hinlink h28k.

amazingfate (approved):

PR #343, opened 2025-03-18
Add NPU to zero3 device tree

halabut69:
Sourced from radxa rk3568-npu-enable.dts


arch/arm64/boot/dts/rockchip/rk3566-radxa-zero3.dtsi
```diff
@@ -453,7 +453,18 @@
 };
 
 &rknpu {
-	rknpu-supply = <&vdd_gpu>;
```
amazingfate, line 456: The overlay from radxa does not delete this line, I think this line is fine.

amazingfate:
default branch is bumpped to rkr5.1, please create a new pr.

PR #345, opened 2025-03-19
arm64: dts: rockchip: add generic rk3568 overlays

### LINKED ISSUE #344: rk3568 standard overlays are missing
Joshua has them. Maybe I port them later or anyone else 
https://github.com/Joshua-Riek/linux-rockchip/commit/d7db06ce0ea471a8decd3e057511e897997c1d15

EvilOlaf:
Fixes https://github.com/armbian/linux-rockchip/issues/344

Borrowed from https://github.com/Joshua-Riek/linux-rockchip/commit/d7db06ce0ea471a8decd3e057511e897997c1d15

Untested


arch/arm64/boot/dts/rockchip/overlay/Makefile
```diff
@@ -200,8 +200,84 @@ dtbo-$(CONFIG_ARCH_ROCKCHIP) += \
...
+	rk3568-gmac1-enable.dtbo \
+	rk3568-i2c1.dtbo \
+	rk3568-i2c2-disable.dtbo \
+	rk3568-i2c2-m0.dtbo \
+	rk3568-i2c2-m0-pca9555.dtbo \
+	rk3568-i2c2-m1.dtbo \
+	rk3568-i2c3-m0.dtbo \
+	rk3568-i2c4-m0.dtbo \
+	rk3568-npu-disable.dtbo \
+	rk3568-npu-enable.dtbo \
```
amazingfate: This should be not necessary, npu is in soc, no reason to enable/disable it.

EvilOlaf: So is it enabled by default then? If so the overlay for disabling could stay. Maybe somebody wants to save some energy or whatever.

amazingfate: npu node is enabled at board level in sdk kernel, but it should be enabled by default.

PR #352, opened 2025-04-21
RK3588 Overlay: Allow DMC to use LPDDR5-6400

HeyMeco:
# Description

This is related to: https://github.com/hbiyik/rkddr which is a tool that allows to easily modify the DDR TPL for Rockchip SoC's. 

Using this overlay with [rkddr](https://github.com/hbiyik/rkddr) tool from @hbiyik we can make use of **LPDDR5-6400** speeds instead of the LPDDR5-5500 set by Rockchip.

# Testing

Radxa Boards:
- 5T (tested with H58G56AK6B-X069, LPDDR5-6400 module)
- 5B+ (tested with H58GG8AK8Q-X103, LPDDR5X-8533 module limited by RK3588 controller)

## Testing method

- Set DMC governor to performance and confirm checking `cur_freq`

# Results

- For GPU related tasks we see an up to **20% uplift** when using glmark2-wayland with panthor driver on the terrain scene

> [!CAUTION]
> RK3588 is validated for up to LPDDR5-5500 any damage caused to your hardware by choosing to enable this overlay as well as setting rkddr speeds is your own fault.

PR #356, opened 2025-04-28
RK3588 Overlay: add dmc-oc-3500mhz overlay to makefile

Sylensky:
This overlay file was actually excluded from the build process and is therein a regression from https://github.com/armbian/linux-rockchip/pull/352.

Tested and verified on a Rock-5T board with the https://github.com/hbiyik/rkddr tool.


nyanmisaka:
@HeyMeco

amazingfate (approved):

PR #357, opened 2025-05-01
add cpu and mem supply to all big cores on rk3582 boards

amazingfate:
A user has reported that his rock5c lite only runs at minimal cpu freq. His rock5c lite has a broken cpu6, which is disabled by uboot(cpu_b2). Now all rk3588 boards has the cpu-supply defined at cpu_b2, when the node is disabled, the cpu freq driver can't be initialized.

He has confirmed that this issue doesn't happend with mainline kernel because all cpu core nodes has defined cpu-supply in mainline devicetree, so I add cpu-supply and mem-supply to the two rk3582 boards.

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3588s-radxa-cm5.dtsi
+++ b/arch/arm64/boot/dts/rockchip/rk3588s-radxa-cm5.dtsi
@@ -132,9 +132,21 @@
 	mem-supply = <&vdd_cpu_big0_mem_s0>;
 };
 
+&cpu_b1 {
+	cpu-supply = <&vdd_cpu_big0_s0>;
+	mem-supply = <&vdd_cpu_big0_mem_s0>;
+	#cooling-cells = <2>;
+};
+
 &cpu_b2 {
 	cpu-supply = <&vdd_cpu_big1_s0>;
 	mem-supply = <&vdd_cpu_big1_mem_s0>;
+};
+
+&cpu_b3 {
+	cpu-supply = <&vdd_cpu_big1_s0>;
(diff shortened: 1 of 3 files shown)


kwankiu:
shouldn't this be applied to rk3588s-radxa-cm5.dtsi too?

amazingfate:
> shouldn't this be applied to rk3588s-radxa-cm5.dtsi too?

radxa cm5 is a rk3588s board which has all 8 cores, so it doesn't have the issue when one big core is disabled by uboot.

kwankiu:
> > shouldn't this be applied to rk3588s-radxa-cm5.dtsi too?
> 
> radxa cm5 is a rk3588s board which has all 8 cores, so it doesn't have the issue when one big core is disabled by uboot.

Radxa CM5 has RK3582 variants, I own a CM5 Lite myself, although this doesnt affect my board as i have an RK3582 that all cpu, gpu and video decoder seems to be good.

amazingfate:
> > > shouldn't this be applied to rk3588s-radxa-cm5.dtsi too?
> > 
> > radxa cm5 is a rk3588s board which has all 8 cores, so it doesn't have the issue when one big core is disabled by uboot.
> 
> Radxa CM5 has RK3582 variants, I own a CM5 Lite myself, although this doesnt affect my board as i have an RK3582 that all cpu, gpu and video decoder seems to be good.

I didn't know it has a lite version. I will add it later.

amazingfate:
I find that pwm-fan also need the cooling-cells property on big core. Missing this property will cause pwm fan always working at full speed.

PR #362, opened 2025-05-27
slow down sdio freq to 100M to armsom rk3576 boards

amazingfate:
It's reported that the current sdio freq will cause wifi unstable, it's better to slow it down to 100M.


chainsx:
LGTM

PR #374, opened 2025-07-17
Sync Panthor with drm-tip (uAPI 1.5)

ginkage:
These weren't merged into 6.16 yet, but there are already some patches proposed for uAPI 1.6, so I wanted to get these out to make the follow-up backport easier. Any testing would be highly appreciated.


amazingfate (approved):

ginkage:
Friendly bump.

PR #380, opened 2025-08-06
Panthor with 1GHz frequency

ginkage:
Kwiboo's patches backported to BSP kernel, though I've only included the rk3588 part for now.


hbiyik:
thoroughly tested and works on my 3588

HeyMeco (approved):

PR #381, opened 2025-08-10
Sync Rock5B, 5B-Plus, 5T DT's

HeyMeco:
From Radxa's Kernel fork: https://github.com/radxa/kernel/blob/linux-6.1-stan-rkr5.1/arch/arm64/boot/dts/rockchip/

Things included are:
- fix usb device disconnect event missing issue
- https://github.com/radxa/kernel/commit/22df35500817a0370e4a8e66f059b983b4668545

Tested with:
1. Rock-5B-Plus

@amazingfate for Rock5B the biggest change includes a simpler `&soc_thermal` node. Should we sync that part with Radxa's kernel or keep your variant?


amazingfate:
soc_thermal node is related to pwm fan cooling levels. I recommend keeping the old one.

PR #401, opened 2025-09-21
Adding Mecotronic R58 HD

igorpecovnik:
Patches were pulled from their repo.

Changes to VOP driver that might affect other boards.


amazingfate:
About the driver change at drm driver, I assume the author is copying code from another version of kernel. A lot of code deletion and rename of variables/functions. That's too bad.

amazingfate:
Here is the rebased drm changes from mecotronic:
https://github.com/armbian/linux-rockchip/commit/23b1d042fff4f788a263f7f96ab4f563088b51ed

It is setting preset edid instead from monitor, I have no idea if the change is necessary.

igorpecovnik:
This is then safe to merge, right? I propose Mekotronics to rework DRM changes or try to used other way.

amazingfate:
> This is then safe to merge, right? I propose Mekotronics to rework DRM changes or try to used other way.

Yes, it's safe now

PR #409, opened 2025-10-21
arch: arm: dts: add Luckfox Pico Mini (RV1103) + rv1106 OP-TEE

vidplace7:
Add first RV1103/RV1106 family DTS. Luckfox Pico Mini.
Taken from Luckfox Linux 5.10 SDK and modified for Armbian.

Also modifies base rv1106.dtsi (which is included by the rv1103.dtsi) to reserve memory for OP-TEE.


EvilOlaf:
At some point we will switch default branch to https://github.com/armbian/linux-rockchip/tree/rk-6.1-rkr6.1 which currently needs some stuff figured out. However main rebase has been done already which means your PR will be lost when the switch happens.
I suggest to PR against this branch too.

amazingfate (approved):

PR #422, opened 2025-11-08
drm/panthor: Flush shmem writes before mapping buffers CPU-uncached

HeyMeco:
From upstream: https://lore.kernel.org/dri-devel/20251107171214.1186299-1-boris.brezillon@collabora.com/T/#u

The shmem layer zeroes out the new pages using cached mappings, and if we don't CPU-flush we might leave dirty cachelines behind, leading to potential data leaks and/or asynchronous buffer corruption when dirty cachelines are evicted.

Fixes: 8a1cc07578bf ("drm/panthor: Add GEM logical block")


HeyMeco:
The patch applied cleanly but needs to be tested to make sure it didn't break anything downstream. 
If confirmed let's get this merged as quick as possible

PR #465, opened 2026-03-30
drm: add support for rpi panel v2

ackPeng:
Added support for the Raspberry Pi v2 7-inch touchscreen; the main code comes from the official Raspberry Pi Linux kernel.
Verification has been performed on 3576 and 3588.

### DIFF
--- a/drivers/gpu/drm/panel/panel-ilitek-ili9881c.c
+++ b/drivers/gpu/drm/panel/panel-ilitek-ili9881c.c
@@ -26,6 +26,11 @@ enum ili9881c_op {
 	ILI9881C_COMMAND,
 };
 
+enum ili9881_desc_flags {
+	ILI9881_FLAGS_NO_SHUTDOWN_CMDS = BIT(0),
+	ILI9881_FLAGS_PANEL_ON_IN_PREPARE = BIT(1),
+};
+
 struct ili9881c_instr {
 	enum ili9881c_op	op;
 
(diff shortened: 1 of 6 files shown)


igorpecovnik (approved):
> Verification has been performed on 3576 and 3588.

OK.

In general we are also concerned that code changes won't break other devices. Where / if you are not sure, add conditions.

ackPeng:
Okay, I will review and modify the code again to ensure it does not negatively impact other platforms.

ackPeng:
This modification should not affect previous platforms:

1. This modification mainly changes the `panel-ilitek-ili9881c.c` and `goodix.c/h` files, and adds the `rpi-panel-v2-regulator.c` file. The newly added `rpi-panel-v2-regulator.c` file will not affect other platforms.

2. For the `panel-ilitek-ili9881c.c` file, support for Raspberry Pi and DSI-7-inch panels has been added. Compared to the previous file, `lanes` and `flags` member variables have been added to `ili9881c_desc`, and the configuration of these two member variables has been made for other panels in the file (the configured values ​​are the same as the previous default configurations). Conditional judgments have been added to functions such as `ili9881c_enable` to ensure that the behavior logic of the previous panels remains unchanged.

3. For `goodix.c`, the original IRQ interrupt reporting path is retained, and a polling mechanism has been added in the absence of IRQ, without affecting the previous IRQ mechanism.

PR #466, opened 2026-04-03
arch: arm: dts: add Luckfox Lyra

netrunnereve:
This adds support for the regular Luckfox Lyra board, it boots and runs fine. See the build PR https://github.com/armbian/build/pull/9630 for more details.

I don't have a way to test the spi changes in https://github.com/armbian/linux-rockchip/pull/450 so unlike the already supported Plus board this one doesn't have spi enabled.


EvilOlaf (approved):

vidplace7 (changes requested):
The existing lyra-plus definition should work fine on lyra. (Simply disable the ethernet if you'd like, `rockchip-rk3506-disable-ethernet0` overlay will do it.)

Is that not the case?

PR #472, opened 2026-04-22
change recomputer rk35xx fan speed

ackPeng:
change recomputer rk35xx fan speed and pca9535 pin name and recomputer rk3588 sdio frequency


SuperKali (approved):
LGTM

PR #474, opened 2026-04-25
libbpf: fix const correctness in resolve_full_path for gcc15

EvilOlaf:
`vendor` header compilation fails when building with `resolute` as target. This is likely due to gcc15 being default in Resolute. I threw this at llm and this seem to fix the compilation. 
This was originally discovered because aic8800 compilation failed due to failing headers.
This also looks quite similar to the mainline upstream fix: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.18.24&id=ab21cf885fb2af179c44d8beeabd716133b9385d

### DIFF
--- a/tools/lib/bpf/libbpf.c
+++ b/tools/lib/bpf/libbpf.c
@@ -10738,7 +10738,7 @@ static int resolve_full_path(const char *file, char *result, size_t result_sz)
 		if (!search_paths[i])
 			continue;
 		for (s = search_paths[i]; s != NULL; s = strchr(s, ':')) {
-			char *next_path;
+			const char *next_path;
 			int seg_len;
 
 			if (s[0] == ':')


amazingfate:
we may rebase to newer rk sdk release, using upstream fix would make developers treat this patch more clearly when bump ing kernel(drop or keep this patch).

EvilOlaf:
I'd leave this at your discretion. Your insights are most likely deeper than mine.

igorpecovnik:
This is probably still good to merge? As I don't expect this rebase to happen soon-ish?

amazingfate:
I just checked there is no related fix in 6.1 lts, and the one for 6.18 in upstream has patched some code not in 6.1 source tree. This PR is fine.

amazingfate (approved):

PR #475, opened 2026-04-30
possibly fix copy.fail (and others?) by picking from 6.1.170

rpardini:
https://copy.fail/

(the payload there is x86-only and won't run on arm64, but the vuln is very much there)

I picked all commits that applied cleanly from 6.1.170, thus probably more than just copy.fail fixed there.

### DIFF
--- a/crypto/Kconfig
+++ b/crypto/Kconfig
@@ -1353,7 +1353,6 @@ config CRYPTO_USER_API_AEAD
 	depends on NET
 	select CRYPTO_AEAD
 	select CRYPTO_SKCIPHER
-	select CRYPTO_NULL
 	select CRYPTO_USER_API
 	help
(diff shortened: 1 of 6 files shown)


rpardini:
Low prio.

Armbian's .config doesn't enable `CONFIG_CRYPTO_USER_API_AEAD`- thus `algif_aead` module is not present -- kernel source is vulnerable, but as shipped by Armbian it is simply not there.

igorpecovnik:
```
   crypto/algif_aead.c: In function '_aead_recvmsg':
   crypto/algif_aead.c:192:9: error: implicit declaration of function 'memcpy_sglist' [-Werror=implicit-function-declaration]
     192 |         memcpy_sglist(rsgl_src, tsgl_src, ctx->aead_assoclen);
         |         ^~~~~~~~~~~~~
```

PR #483, opened 2026-05-18
arm64: dts: rockchip: fix adc-keys press-threshold-microvolt for recomputer devkits

mingzhangqun:
## Summary
- Fix volumeup-key `press-threshold-microvolt` value from 1750 to 1750000 (1.75V) for recomputer rk3576 and rk3588 devkit boards
- The original value was off by a factor of 1000, causing the ADC key detection to malfunction

## Test plan
- [ ] Verify adc-keys node reports correct voltage threshold in sysfs
- [ ] Test volume up key functionality on recomputer rk3576-devkit
- [ ] Test volume up key functionality on recomputer rk3588-devkit


igorpecovnik (approved):

SuperKali (approved):
LGTM

PR #484, opened 2026-05-18
arm64: dts: rockchip: rk3576: remove empty rockchip,supported-hw from OPP tables

mingzhangqun:
## Summary
- Remove empty `rockchip,supported-hw;` property from 7 OPP tables in rk3576.dtsi (cluster0, cluster1, bus_a72, dmc, npu, gpu, vop)
- The bare property with no value makes the OPP framework require `opp-supported-hw` on every OPP entry, but only `opp-j-m-*` entries have it. This causes all generic OPP entries to be rejected, resulting in GPU stuck at ~100 MHz without devfreq

## Test plan
- [ ] Verify GPU devfreq scales properly on rk3576 (non-J/M variant)
- [ ] Check CPU frequency scaling works on cluster0/cluster1
- [ ] Verify no regression on J/M variants that use opp-supported-hw entries

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3576.dtsi
+++ b/arch/arm64/boot/dts/rockchip/rk3576.dtsi
@@ -398,7 +398,6 @@
 
 		nvmem-cells = <&cpul_leakage>, <&cpul_opp_info>, <&specification_serial_number>;
 		nvmem-cell-names = "leakage", "opp-info", "specification_serial_number";
-		rockchip,supported-hw;
 
 		rockchip,pvtm-hw = <0x06>;
 		rockchip,pvtm-voltage-sel-hw = <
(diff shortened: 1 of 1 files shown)


igorpecovnik (approved):

SuperKali:
@mingzhangqun the GPU fix is correct, but the other 6 tables regress.

The bool `rockchip,supported-hw;` is the only trigger that populates `info->supported_hw[]` in `rockchip_set_opp_supported_hw()` (opp_select.c:1452). Without it, `opp_table->supported_hw` stays NULL, and `_opp_is_supported()` (of.c:539-540) rejects every OPP that has `opp-supported-hw`.

Entry count in rk3576.dtsi: cluster0=19/19, cluster1=21/21, bus_a72=6/6, dmc=8/8, npu=16/16, vop=6/6 entries have `opp-supported-hw`. Only gpu has 7/13 without it (after commit e8448b91 removed them for panfrost). rk3588s.dtsi uses the same bool+property pattern and works.

Suggest narrowing the diff to just `gpu_opp_table`.

mingzhangqun:
@SuperKali  Thanks, fixed. The update now narrows the change to gpu_opp_table only, and the other OPP tables keep rockchip,supported-hw.

SuperKali:
Thanks @mingzhangqun, the narrowed diff looks correct. Merging.

PR #495, opened 2026-06-01
drm: rockchip: DSI fixes for RK3576 dual-display

mingzhangqun:
## Summary

Two fixes for RK3576 MIPI/DSI output when used alongside other displays.

### Patches

1. **drm/rockchip: rk3576: switch DSI dclk_src from vpll to cpll** — When both DP and DSI are active, DP reprograms vpll which breaks DSI pixel clock. Switch DSI VP's dclk parent to cpll (1000MHz, supports 720x1280@60 exactly).

2. **drm/rockchip: dw-mipi-dsi2: set panel orientation before device registration** — Fixes  when DSI panel driver calls  from  after DRM device registration.

## Test plan

- [x] Simultaneous HDMI + DP + DSI output works without pixel clock issues
- [x] DSI panel (ili9881c) with portrait orientation: no kernel WARNING
- [x] No regressions on single-display configurations


HeyMeco:
Thank you. Merging this!

PR #496, opened 2026-06-01
net: r8169: add device tree based LED configuration for r8125

mingzhangqun:
## Summary

Read r8125 LED register/value pairs from the PCIe controller device tree `realtek,led-data` property, instead of hardcoding in the shared driver path.

The driver searches DT nodes for `realtek,led-data` and writes each reg/value pair via `RTL_W16`.

### Tested on Seeed reComputer RK3588 DevKit

- [x] Both r8125 Ethernet ports: LED registers correctly written
- [x] No impact on other boards (property absent = default behavior)

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3588-recomputer-rk3588-devkit.dts
+++ b/arch/arm64/boot/dts/rockchip/rk3588-recomputer-rk3588-devkit.dts
@@ -893,6 +893,11 @@
 /* phy1 - right ethernet port PCIE20_1_TXP */
 &pcie2x1l0 {
 	reset-gpios = <&gpio4 RK_PA5 GPIO_ACTIVE_HIGH>;
+	/* r8125 Ethernet LED configuration:
+	 * LED0 (yellow): blink on traffic at all link speeds
+	 * LED3 (green): solid on at all link speeds
+	 */
+	realtek,led-data = <0x18 0x022B 0x96 0x002B>;
 	vpcie3v3-supply = <&vcc5v0_usb_eth>;
 	status = "okay";
 };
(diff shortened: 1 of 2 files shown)


HeyMeco:
This is a better approach using device tree for board-specific LED config. A few suggestions to improve efficiency and safety:

### 1. **Avoid Repeated Memory Allocation**
The function allocates and parses on every call (WoL + hardware init). Cache the config once during probe instead:

```c
// In rtl8169_private struct:
u32 *led_config;
int led_config_count;

// Parse once during probe, then reuse in rtl8125_apply_dt_led_config()
```

This eliminates the `kmalloc_array()` call from the hot path.

### 2. **Eliminate Redundant Calls**
Currently called twice during init—from both `rtl_wol_enable_rx()` and `rtl_hw_start_8125_common()`. Call it once from a single location like `rtl_hw_start_8125()`.

### 3. **Improve Type Safety (Minor)**
```c
for (i = 0; i < count; i += 2)
    RTL_W16(tp, (unsigned int)led_data[i], (u16)led_data[i + 1]);
```

Use `unsigned int` for register addresses instead of `int` cast.

---

mingzhangqun:
@HeyMeco Thanks for the review! All three suggestions applied in the latest push:

1. **Cache config at probe** — `struct rtl8169_private` now holds `led_data` / `led_data_count`. DT parsing (`rtl8125_parse_dt_led_config()`) runs once from `rtl_init_one()`, and `rtl_remove_one()` frees the buffer. The hot-path `rtl8125_apply_dt_led_config()` only iterates the cached array — no more `kmalloc_array()` per call.

2. **Single call site** — removed the call from `rtl_wol_enable_rx()`. LED config is now applied only from `rtl_hw_start_8125_common()`.

3. **Type safety** — dropped the `(int)` cast; register addresses flow as `u32` straight into `RTL_W16()`.

Patch updated to v2 (21bb15e). The diff vs the previous revision is essentially the parse/apply split + the struct fields + the probe/remove wiring.

mingzhangqun:
Superseded by #497 (branch recreated due to CI runner failure). All discussion migrated.

PR #501, opened 2026-06-12
auxdisplay: vk2c21: add sysfs brightness control

rpardini:
#### auxdisplay: vk2c21: add sysfs brightness control

> Expose a `brightness` sysfs attribute that drives the chip's IVASET register (0x8A):

```
  - "default" (the boot-time state): IVASET=0x0F, IVA off, bias driven
    by the board's external VR resistor (same as before this change).
  - 0..15:                          IVASET=0x3F..0x30 (IVA on, SEG mode);
                                    15 = brightest (1.000 x VDD), 0 =
                                    dimmest (0.529 x VDD at 1/3 bias).
```

SEG mode (0x30..0x3F) is used rather than VLCD-output mode (0x10..0x1F)
because the R58X-Pro ties the VLCD pin to VDD through an external
resistor: in VLCD-output mode the external circuit dominates the chip's
internal driver and the brightness register has no visible effect. In
SEG mode the chip ignores the VLCD pin and generates the bias voltage
purely internally, so the 16 levels actually take effect.

Probe-time behaviour is unchanged: init still sends IVA=0x0F, and the
attribute reads back as "default" until userspace writes a number.

### DIFF
--- a/drivers/auxdisplay/lcd-vk2c21.c
+++ b/drivers/auxdisplay/lcd-vk2c21.c
@@ -41,7 +41,23 @@
 #define VK2C21_SYS_OFF_LCD_OFF	0x00	/* System + LCD off */
 #define VK2C21_FRAME_80HZ	0x00	/* 80 Hz frame rate */
 #define VK2C21_BLINK_OFF	0x00	/* Blinking off     */
-#define VK2C21_IVA_DEFAULT	0x0F	/* VLCD selected, IVA off, R1 */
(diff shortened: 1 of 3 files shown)


rpardini:
Tested on-hardware, actually works

rpardini:
Ooops, no, pushed stuff to wrong branch. Will cleanup in a bit

rpardini:
I messed up this PR with non-relevant commits. The (relevant, lol) vk2c21 brightness stuff landed on rkr7.2 already. Closing this.

PR #504, opened 2026-06-14
ata: ahci: force 32-bit DMA for JMicron JMB582/JMB585

HeyMeco:
The JMicron JMB585 (and JMB582) SATA controllers advertise 64-bit DMA support via the S64A bit in the AHCI CAP register, but their 64-bit DMA implementation is defective. Under sustained I/O, DMA transfers targeting addresses above 4GB silently corrupt data -- writes land at incorrect memory addresses with no errors logged.

Add device-specific PCI ID entries for the JMB582 (0x0582) and JMB585 (0x0585) before the generic JMicron class match, using a new board type that combines AHCI_HFLAG_IGN_IRQ_IF_ERR (preserving existing behavior) with AHCI_HFLAG_32BIT_ONLY to force 32-bit DMA masks.

Backport from upstream commit 105c42566a550 ("ata: ahci: force 32-bit DMA for JMicron JMB582/JMB585").

---

This bug is present on RK3588 as it reports 64 bit wide dma.

### DIFF
--- a/drivers/ata/ahci.c
+++ b/drivers/ata/ahci.c
@@ -61,6 +61,7 @@ enum board_ids {
 	/* board IDs for specific chipsets in alphabetical order */
 	board_ahci_al,
 	board_ahci_avn,
+	board_ahci_jmb585,
 	board_ahci_mcp65,
 	board_ahci_mcp77,
 	board_ahci_mcp89,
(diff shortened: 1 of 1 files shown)


HeyMeco:
@igorpecovnik when this is merged can we issue a kernel update via apt?

SuperKali (approved):

x13-me (approved):
lgtm, no concerns

PR #511, opened 2026-07-08
Add kernel level support for KICKPI K3B

retro98boy:
This PR adds kernel-level support for the RK3562 SBC KICKPI K3B. It mainly accomplishes the following:

1.Adds the base DTS for KICKPI K3B.

2.Adds a JD9366-based DSI panel DTB overlay for KICKPI K3B. Since the KICKPI K3B only features a DSI display interface, using a DTBO to support custom panels is a practical approach for users. The screen DTSO provided here is configured for my own panel and can be used as a reference.

3.Adds the I2C touch driver for JD9366. This is the TDDI touch driver for the panel I used, with the source code originating from JADARD.

4.Adds Panfrost support for the RK3562 Mali-G52 GPU. If the clk_gpu, clk_gpu_brg, and aclk_gpu clocks (declared in rk3562.dtsi) are not enabled, the kernel will hang during Panfrost initialization:

```
root@kickpi-k3b:~# modprobe panfrost
[ 2817.332173] panfrost ff320000.gpu: clock rate = 500000000
[ 2817.344151] panfrost ff320000.gpu: EM: created perf domain    <---- Stuck here
```

5.Adds the driver for the MAXIO PHY used by KICKPI K3B.

6.Introduces the `initial-on-delay-ms` property for gpio-leds. This property forces a specified delay before turning on LEDs that are configured to be "on" during the boot process. This is introduced to resolve a specific hardware quirk on the KICKPI K3B:
If the battery functionality is enabled in software but no physical battery is connected, powering the device via Type-C will cause it to hang during the boot process. To workaround this issue, KICKPI introduces a 10-second delay before enabling the USB Type-A power supply.

### DIFF
--- a/arch/arm64/boot/dts/rockchip/Makefile
+++ b/arch/arm64/boot/dts/rockchip/Makefile
@@ -119,6 +119,7 @@ dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3528-rock-2a.dtb
 dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3528-rock-2f.dtb
 dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3528-nanopi-rev01.dtb
 dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3562-dictpen-test3-v20.dtb
+dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3562-kickpi-k3b.dtb
 dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3562-evb1-lp4x-v10.dtb
 dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3562-evb1-lp4x-v10-linux.dtb
 dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3562-evb1-lp4x-v10-linux-amp.dtb
--- a/arch/arm64/boot/dts/rockchip/overlay/Makefile
+++ b/arch/arm64/boot/dts/rockchip/overlay/Makefile
@@ -139,6 +139,7 @@ dtbo-$(CONFIG_ARCH_ROCKCHIP) += \
 	rockchip-rk3588-panthor-gpu.dtbo \
 	rk3528-rock-2-enable-pcie.dtbo \
 	rk3566-roc-pc-sata2.dtbo \
+	rk3562-kickpi-k3b-demo-dsi-panel.dtbo \
 	rk3576-can1-m1.dtbo \
 	rk3576-i2c0-m1.dtbo \
 	rk3576-i2c1-m0.dtbo \
--- a/arch/arm64/boot/dts/rockchip/overlay/rk3562-kickpi-k3b-demo-dsi-panel.dts
+++ b/arch/arm64/boot/dts/rockchip/overlay/rk3562-kickpi-k3b-demo-dsi-panel.dts
@@ -1,0 +1,257 @@
+/dts-v1/;
+/plugin/;
+
+/*
+ * Simple reference DSI panel DT overlay.
+ * Works with JD9366-based 800x1280 panels.
+ * Verified on SQ101P-X4EI451-84H501 and YDP1010BT003-V1.
+ */
+
+#include <dt-bindings/display/drm_mipi_dsi.h>
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/interrupt-controller/irq.h>
+#include <dt-bindings/pinctrl/rockchip.h>
+
+/ {
+	fragment@0 {
+		target-path = "/";
+
+		__overlay__ {
+			lcd_3v3: regulator-lcd-3v3 {
+				compatible = "regulator-fixed";
+				enable-active-high;
+				gpio = <&gpio0 RK_PB0 GPIO_ACTIVE_HIGH>;
+				pinctrl-names = "default";
+				pinctrl-0 = <&lcd_3v3_pin>;
+				regulator-min-microvolt = <3300000>;
+				regulator-max-microvolt = <3300000>;
+				regulator-name = "lcd_3v3";
+				vin-supply = <&vcc_3v3>;
+			};
+
+			backlight_dsi: backlight-dsi {
+				compatible = "gpio-backlight";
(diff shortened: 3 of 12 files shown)


HeyMeco:
In general this looks good but I'm not quite able to follow why the panfrost changes were needed as that GPU has been supported with Rockchip SoC for a while? 
Can’t you replicate the gpu dt like on RK35-66/68/76 to make it work?

retro98boy:
> In general this looks good but I'm not quite able to follow why the panfrost changes were needed as that GPU has been supported with Rockchip SoC for a while? Can’t you replicate the gpu dt like on RK35-66/68/76 to make it work?

The Mali GPU requires specific clocks to function properly. For the Panfrost driver, its clock request logic is handled as follows:

```c
// https://github.com/armbian/linux-rockchip/blob/rk-6.1-rkr5.1/drivers/gpu/drm/panfrost/panfrost_device.c

static int panfrost_clk_init(struct panfrost_device *pfdev)
{
	int err;
	unsigned long rate;

	pfdev->clock = devm_clk_get(pfdev->dev, NULL);
	if (IS_ERR(pfdev->clock)) {
		dev_err(pfdev->dev, "get clock failed %ld\n", PTR_ERR(pfdev->clock));
		return PTR_ERR(pfdev->clock);
	}

	rate = clk_get_rate(pfdev->clock);
	dev_info(pfdev->dev, "clock rate = %lu\n", rate);

	err = clk_prepare_enable(pfdev->clock);
	if (err)
		return err;

	pfdev->bus_clock = devm_clk_get_optional(pfdev->dev, "bus");
	if (IS_ERR(pfdev->bus_clock)) {
		dev_err(pfdev->dev, "get bus_clock failed %ld\n",
			PTR_ERR(pfdev->bus_clock));
		err = PTR_ERR(pfdev->bus_clock);
		goto disable_clock;
	}

	if (pfdev->bus_clock) {
		rate = clk_get_rate(pfdev->bus_clock);
		dev_info(pfdev->dev, "bus_clock rate = %lu\n", rate);

		err = clk_prepare_enable(pfdev->bus_clock);
		if (err)
			goto disable_clock;
	}

	return 0;

disable_clock:
	clk_disable_unprepare(pfdev->clock);

	return err;
}
```

As shown in the driver source, Panfrost requests the first clock defined in the DTS (regardless of its name), and then requests a second clock explicitly named "bus".

For RK3568/RK3566 (rk356x.dtsi), the clocks in the GPU node are defined like this:

```
clocks = <&scmi_clk 1>, <&cru CLK_GPU>;
clock-names = "gpu", "bus";
```

This works perfectly out-of-the-box for RK3568/RK3566. All that's required to make Panfrost work there is converting the interrupt-names to lowercase ("gpu", "mmu", "job"), which is a Panfrost requirement.

However, for RK3576, the GPU clocks are defined differently:

```
clocks = <&scmi_clk CLK_GPU>, <&cru CLK_GPU>;
clock-names = "clk_mali", "clk_gpu";
```

Obviously, Panfrost cannot obtain the clk_gpu clock using this configuration. Proceeding with subsequent operations—such as accessing GPU registers without this clock enabled—leads to severe consequences, such as an SoC hang/freeze.

Fortunately, @amazingfate used a clever [commit](https://github.com/armbian/linux-rockchip/commit/d6cac1d7d47713f2efc78a228c006ad7c05ff75f) to enable clk_gpu without modifying the Panfrost driver itself: by handing over clk_gpu to the power-domain to handle its request and enablement. This leaves Panfrost responsible only for requesting the first clock (which is likely the core clock that scales frequency up/down for DVFS).

I initially attempted to mimic this approach for RK3562, but the Panfrost probe timed out and exited. For reference, here are the required GPU clocks on the RK3562:

```
clocks = <&scmi_clk CLK_GPU>, <&cru CLK_GPU>,
         <&cru CLK_GPU_BRG>, <&cru ACLK_GPU_PRE>;
clock-names = "clk_mali", "clk_gpu", "clk_gpu_brg", "aclk_gpu";
```

I plan to experiment further with modifying the DTS structure to properly map these clock requests. If that doesn't yield a solution, I will report back here. Realistically, adding a specific patch to Panfrost shouldn't impact other SoCs, though it might look a bit less elegant.

@amazingfate, do you have any suggestions or insights on this? Any guidance would be highly appreciated!

retro98boy:
@HeyMeco I spent a long time today trying to achieve a successful Panfrost load on the RK3562 solely by modifying the DTS, but none of my attempts succeeded. Every probe ended up causing an SoC hang. Therefore, I believe modifying the Panfrost source code to enable these clocks is the only way forward.

retro98boy:
> [retro98boy](https://github.com/retro98boy) [force-pushed](https://github.com/armbian/linux-rockchip/compare/1340d5c19f13f1311a0428252a2e4c6c0f3872ba..e5af084fc321a02628f227ba604f66803867c6cd) the kickpi-k3b branch from [1340d5c](https://github.com/armbian/linux-rockchip/commit/1340d5c19f13f1311a0428252a2e4c6c0f3872ba) to [e5af084](https://github.com/armbian/linux-rockchip/commit/e5af084fc321a02628f227ba604f66803867c6cd) [12 minutes ago](https://github.com/armbian/linux-rockchip/pull/511#event-28846545384)

just rebase against latest HEAD

retro98boy:
PR V2:

- Rename `kickpi-k3b-sq101p-x4ei451-84h501.dtbo` to `rk3562-kickpi-k3b-demo-dsi-panel.dtbo` as a minimal example. It contains no I2C touch node and no PWM dimming table (backlight is always on or off).

- Remove JADARD JD9366T touchscreen driver (Users who actually need this can download the source code and build the kernel module themselves)

- Refactor the RK3562-specific Panfrost extra clock patch to a more generic implementation. Instead of hardcoding RK3562 clock names in the .c file, retrieve them directly from the Device Tree (DTS).

retro98boy:
@HeyMeco Hi, I'd really appreciate it if you could take another look at this PR when you have time to see if it meets the requirements.

To ensure rk3562 panfrost probes correctly, I modified the panfrost driver, but made the implementation more generic. And it shouldn't break panfrost functionality on other SoCs.

Additionally, to avoid maintenance overhead from introducing uncommon drivers, I removed the jd9366t driver code.

PR #520, opened 2026-08-07
rkr7.2: fix 32-bit build (`arm: dts: drop rv1126b-evb1-v12-4x-cam.dts`)

rpardini:
#### rkr7.2: fix 32-bit build (`arm: dts: drop rv1126b-evb1-v12-4x-cam.dts`)

### DIFF
--- a/arch/arm/boot/dts/Makefile
+++ b/arch/arm/boot/dts/Makefile
@@ -1199,7 +1199,6 @@ dtb-$(CONFIG_ARCH_ROCKCHIP) += \
 	rv1126b-evb1-v12-spi-nand.dtb \
 	rv1126b-evb1-v12-uvc.dtb \
 	rv1126b-evb1-v12-uvc-winhello.dtb \
-	rv1126b-evb1-v12-4x-cam.dts \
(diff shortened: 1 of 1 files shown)


arch/arm/boot/dts/Makefile
```diff
@@ -1199,7 +1199,6 @@ dtb-$(CONFIG_ARCH_ROCKCHIP) += \
 	rv1126b-evb1-v12-spi-nand.dtb \
 	rv1126b-evb1-v12-uvc.dtb \
 	rv1126b-evb1-v12-uvc-winhello.dtb \
-	rv1126b-evb1-v12-4x-cam.dts \
```
nyanmisaka, line 1202: Or perhaps it would be better to revert 18d0ea831815b38663883e2f9b58502da11e7faa? This avoids leaving an orphan file: `arch/arm/boot/dts/rv1126b-evb1-v12-4x-cam.dts`

rpardini: Good catch. CI actually built `armhf` OK with the simple Makefile removal, I'll rework as a revert.

PR #503, opened 2026-06-13
input: remotectl: rockchip-pwm: prefer dedicated channel-3 IRQ

rpardini:
On PWM v4 (RK3588), channel 3 of each PWM controller block exposes two interrupts: a group IRQ at index 0 shared with channels 0/1/2 of the same block, and a dedicated capture IRQ at index 1 intended for IR receive / power-key wakeup use of channel 3.

The driver always grabbed index 0 and requested it without IRQF_SHARED, which collides with pwm-rockchip when another channel in the same block is enabled (e.g. pwm1 for a fan on the FriendlyElec CM3588-NAS). The in-tree v4 path in pwm-rockchip.c also omits IRQF_SHARED, so probe fails with -EBUSY:

  remotectl-pwm fd8b0030.pwm: cannot claim IRQ 28
  remotectl-pwm: probe of fd8b0030.pwm failed with error -16

Prefer index 1 when present and fall back to index 0 for PWM v1-v3 nodes that only expose a single interrupt. The capture-mode handler (rockchip_pwm_irq_v4) is unchanged and remains correct for the dedicated IRQ.

PR #523, opened 2026-08-15
input: rockchip_pwm_remotectl: fix IRQ selection and sharing on PWM v1

sormy:
Follow-up to #503, which is correct on v4 but not on v1.

**Wrong IRQ.** The code takes `platform_get_irq(pdev, 1)` and falls back to index 0
on failure. On v4 index 1 is channel 3's dedicated capture IRQ; on v1 it is the
pwrmatch IRQ and capture lives on the group IRQ at index 0. The lookup succeeds on
v1, so the fallback never runs — `irq 28: nobody cared`. Select by `pwm_version`;
the driver defines only 1 and 4.

**Non-shared request.** On v1 the group IRQ is shared with the other channels, and
`pwm-rockchip` claims it `IRQF_NO_SUSPEND | IRQF_SHARED` for `main_version < 4`, so
a non-shared request is refused. Both handlers already read `PWM_REG_INTSTS` and
return `IRQ_NONE`, so sharing is safe.

**Wake source.** Moving capture to index 0 also moves what `enable_irq_wake()` arms,
but the power key arrives on pwrmatch — `remotectl_suspend()` masks the capture
channel and leaves `PWM_PWR_INT_ENABLE` set. `pwr_irq` is now marked too.

Verified on an RK3528 box:

```
irq 28  rk_pwm_irq ×4    wakeup=enabled   (group, shared with pwm-rockchip)
irq 57  rk_pwm_pwr_irq   wakeup=enabled   (pwrmatch)
```

Receiver enumerates as `event2`, every remote button produces key events, the power
key cold-boots from off, no `nobody cared` and no flags mismatch. Not covered:
resume from s2idle — deep-suspend wake on this SoC goes through PSCI/PMU. No change
on v4.

### DIFF
--- a/drivers/input/remotectl/rockchip_pwm_remotectl.c
+++ b/drivers/input/remotectl/rockchip_pwm_remotectl.c
@@ -513,11 +513,19 @@ static int rockchip_pwm_set_pwrmatch(struct platform_device *pdev)
 		goto end;
 	}
 	ret = devm_request_irq(&pdev->dev, pwr_irq, rockchip_pwm_pwrirq,
-			IRQF_NO_SUSPEND, "rk_pwm_pwr_irq", ddata);
+			IRQF_NO_SUSPEND | IRQF_SHARED, "rk_pwm_pwr_irq", ddata);
 	if (ret) {
 		dev_err(&pdev->dev, "cannot claim PWR_IRQ!!!\n");
 		goto end;
 	}
+	/*
+	 * remotectl_suspend() masks the capture channel and leaves
+	 * PWM_PWR_INT_ENABLE set, so the power key arrives here rather than on
+	 * the capture IRQ that probe arms for wake. IRQF_NO_SUSPEND keeps this
+	 * one delivered but does not make it a wake source, so mark it too.
+	 */
+	if (enable_irq_wake(pwr_irq))
+		dev_warn(&pdev->dev, "cannot set PWR_IRQ as wake source\n");
 	val = readl_relaxed(ddata->base + PWM_REG_INT_EN(pwm_id));
 	val = (val & 0xFFFFFF7F) | PWM_PWR_INT_ENABLE;
 	writel_relaxed(val, ddata->base + PWM_REG_INT_EN(pwm_id));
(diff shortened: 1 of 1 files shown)


sormy:
Good finding — confirmed and fixed in 815e609.

You're right that this patch moves what gets armed: `enable_irq_wake()` is
applied to the selected capture IRQ, so switching v1 from index 1 to index 0
moved wake arming off the line the power key actually uses.
`remotectl_suspend()` masks the capture channel and leaves `PWM_PWR_INT_ENABLE`
set, and `IRQF_NO_SUSPEND` keeps that line delivered without making it a wake
source. `pwr_irq` is now marked as a wake source, with the return checked.

Two caveats I'd rather state than have you find:

- Not yet exercised across a suspend cycle. On this SoC deep-suspend wake runs
  through PSCI/PMU (`remote_support_psci = 1`, `rockchip-suspend` present), which
  is why the board woke regardless; the arming matters for s2idle and for boards
  without that path.
- `rk_pwm_remove()` is empty, so the pre-existing `enable_irq_wake()` was already
  unbalanced on unbind. This follows the same pattern rather than reworking the
  driver's wake lifecycle in a fix about IRQ selection.

sormy:
Closed the loop on this without needing a suspend cycle — `/sys/kernel/irq/*/wakeup`
exposes the genirq wake state directly (`CONFIG_GENERIC_IRQ_SYSFS`), so the arming
is readable. Rebuilt and rebooted with the fix:

```
irq 28  rk_pwm_irq,rk_pwm_irq,rk_pwm_irq,rk_pwm_irq   wakeup=enabled   (hwirq 85, capture)
irq 57  rk_pwm_pwr_irq                                wakeup=enabled   (hwirq 86, pwrmatch)
```

`irq` now resolves to index 0, so the only path that arms 57 is the added
`enable_irq_wake(pwr_irq)`; it returns success too, as the new `dev_warn` never
fires. No `nobody cared`, no flags mismatch.

Still not covered, and the PR says so: whether a power-key press actually resumes
from s2idle. That is beyond the arming itself — on this SoC deep-suspend wake runs
through PSCI/PMU, not the IRQ wake framework.

PR #544, opened 2026-09-04
arm64: dts: rockchip: mixtile-edge2: fix Bluetooth wake GPIOs and enable eMMC HS400ES

evtest-hash:
Two device tree defects on the Mixtile Edge 2.

I work at Focalcrest/Mixtile and have the EDGE2 V1.2.1 and
Mixtile-Core-3568 V1.2.1 schematics; both changes come off those.

Verified on an Edge 2 running 6.1.115 built from this branch with
`./compile.sh kernel BOARD=mixtile-edge2 BRANCH=vendor` on
armbian/build v25.11, A/B'd by swapping only the DTB on the same kernel:

- BT: the wake interrupt moves from GPIO3_A2 to GPIO3_A1, where the
  module actually drives it. The three bt_* pinctrl groups also get
  applied for the first time.
- eMMC: timing spec 9 (HS200) -> 10 (HS400 ES). fio 3.41 with O_DIRECT,
  read-only on /dev/mmcblk0: sequential 1M qd8 168 -> 270 MiB/s;
  random 4k qd32 6502 -> 6978 IOPS, i.e. unchanged as expected.

Scope: this PR is only these two. Further fixes for this board (PCIe
supply naming, USB hub reset and M.2 card power modelling, Type-C USB
2.0 capping, unused-node removal) will follow separately, and I'll
mirror this set to rk-6.1-rkr7.2 once it lands here.

### DIFF
--- a/arch/arm64/boot/dts/rockchip/rk3568-mixtile-edge2.dts
+++ b/arch/arm64/boot/dts/rockchip/rk3568-mixtile-edge2.dts
@@ -994,6 +994,8 @@
 	supports-emmc;
 	non-removable;
 	max-frequency = <200000000>;
+	mmc-hs400-1_8v;
+	mmc-hs400-enhanced-strobe;
 	status = "okay";
 };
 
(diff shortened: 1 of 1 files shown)


rpardini (approved):
NIce, thanks. I've sent [a fix Claude found when I was testing rkr7.2](https://github.com/armbian/linux-rockchip/pull/540/changes/02d5029d7798658f9a2bb7b8841f0c464b6a75b8), could you please check it as well?

Also, do we have any schematics available for the edge2? I couldn't find it anywhere...

PR #548, opened 2026-09-09
net: stmmac: fix dead Ethernet after suspend on RK3588 (backport the 6.15 phylink/stmmac resume ordering)

### LINKED ISSUE #547: stmmac: Ethernet link dead after suspend/resume on RK3588 (phylink started before the MAC soft reset) — upstream fix not in rk-6.1
**Kernel:** rk-6.1-rkr7.2 and rk-6.1-rkr5.1 (both carry the affected ordering; so does linux-6.1.y). Measured on 6.1.75 (same code), Orange Pi 5B (RK3588S, gmac1 + Motorcomm YT8531), Ubuntu 24.04 userspace.

### Symptom
Suspend to RAM (`rtcwake -m mem`, or systemd suspend) for more than ~1 minute, wake up: everything resumes except wired Ethernet. `ethtool` says "Link detected: no", `ip link set down/up` and `ethtool -r` don't help, unbind/rebind of the GMAC usually does, power-cycle always does. Short sleeps survive. Stock resume path: **11 of 21 sleeps ≥ 90 s came back with a dead link**, 0 of 9 shorter ones.

### Root cause
`stmmac_resume()` calls `phylink_resume()` and only then `stmmac_hw_setup()`, which soft-resets the MAC (`stmmac_init_dma_engine()` → `stmmac_reset()`). Starting phylink first lets the PHY state machine run `genphy_config_aneg()` from a workqueue concurrently with that reset. Its read-modify-write of `MII_ADVERTISE` can hit the MAC's MDIO block mid-reset: the read returns 0 and the register is written back as **`0x0de0`** — phylib's own bits with the two bits it preserves (bit 0, the IEEE selector; bit 12) cleared. Selector 0 never completes autonegotiation, so the link stays down until the PHY is hard-reset.

Proof from userspace, in the dead state: `ANAR` reads `0x0de0`; writing `0x1de1` and restarting autoneg brings the link up within 12 s (4/4). Clearing the PHY's auto-sleep, rewriting its extended registers, `ethtool -r`, link down/up: no effect. `ethtool -s … autoneg on` cannot repair it because phylib preserves bit 0 on every write.

### Upstream
Fixed in 6.15 by Russell King:
- 367f1854d442 `net: phylink: add phylink_prepare_resume()`
- ef43e513 `net: stmmac: address non-LPI resume failures properly` — moves `phylink_resume()` after the hardware setup ("it may bring the link up immediately in a workqueue thread, which will race with initialisation") and uses `phylink_prepare_resume()` to keep the PHY's RX clock running for the DMA reset (the reason the early call existed: 36d18b5664ef).
Not in linux-6.1.y (new phylink API).

### Backport (three patches, apply on rk-6.1-rkr7.2 — PR to follow if wanted)
1. `net: phylink: add phylink_prepare_resume()` (+ `mac_requires_rxc` flag)
2. `net: stmmac: resume the PHY before, and start phylink after, the MAC setup` (adapted to 6.1's two-branch call)
3. `net: phy: motorcomm: YT8531 - redo the init on resume` — second, smaller bug: stmmac hard-resets the PHY on every resume (`stmmac_mdio_reset()`) and with MAC-managed PM nothing re-runs `config_init()`, so the YT8531's extended registers (0xa010/0xa012: drive strength, clock out) sit at power-on defaults after the first suspend.

### Result
With the three patches: **7 of 7 wakes clean** (116 s … 10 min), link up on its own within 3 s at 1 Gbps, `ANAR = 0x1de1` at every wake. The identical six-cycle run without them: 4 of 6 dead.

### Stopgap without a rebuild
`/usr/lib/systemd/system-sleep/` script: on `post`, `echo fe1c0000.ethernet > /sys/bus/platform/drivers/rk_gmac-dwmac/unbind` then `bind` (gmac0: `fe1b0000.ethernet`) — the re-probe hard-resets the PHY; costs a DHCP renew.

defcom5-rockchip:
Fixes #547.

After a suspend to RAM of more than about a minute, wired Ethernet on RK3588 boards often never comes back (link stays down until the PHY is hard-reset). `stmmac_resume()` starts phylink before `stmmac_hw_setup()` soft-resets the MAC, so the PHY state machine's first `MII_ADVERTISE` read-modify-write can hit the MDIO block mid-reset: the read returns 0 and the advertisement is written back as `0x0de0` (selector field cleared), and autonegotiation never completes. Details and the userspace proof are in #547.

This is the upstream fix from 6.15 backported to the 6.1 vendor tree:

1. **net: phylink: add phylink_prepare_resume()** — backport of 367f1854d442 plus the `mac_requires_rxc` config flag it keys on.
2. **net: stmmac: resume the PHY before, and start phylink after, the MAC setup** — backport of ef43e513, adapted to 6.1's two-branch `phylink_resume()` call: `phylink_prepare_resume()` keeps the PHY's RX clock running for the DMA reset (the reason the early `phylink_resume()` existed, 36d18b5664ef), and `phylink_resume()` moves after the hardware initialisation.
3. **net: phy: motorcomm: YT8531 - redo the init on resume** — found on the way: stmmac hard-resets the PHY on every resume (`stmmac_mdio_reset()`) and, with MAC-managed PM, nothing re-runs `config_init()`, so the YT8531's extended registers (0xa010/0xa012) sit at power-on defaults after the first suspend. Independent of the race; harmless to take separately.

Measured on an Orange Pi 5B (RK3588S, gmac1 + YT8531), same code as rkr7.2 in these paths:
- before: 11 of 21 sleeps ≥ 90 s woke with a dead link (`ANAR = 0x0de0`)
- after: 7 of 7 wakes clean (116 s to 10 min), link up on its own within 3 s at 1 Gbps, `ANAR = 0x1de1`

Applies cleanly on `rk-6.1-rkr7.2` (and rkr5.1). Compile-tested on 6.1.75 with an RK3588 config, no new warnings.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/drivers/net/ethernet/stmicro/stmmac/stmmac_main.c
+++ b/drivers/net/ethernet/stmicro/stmmac/stmmac_main.c
@@ -1260,6 +1260,8 @@ static int stmmac_phy_setup(struct stmmac_priv *priv)
 		priv->phylink_config.mac_capabilities &=
 			~(MAC_10HD | MAC_100HD | MAC_1000HD);
 	priv->phylink_config.mac_managed_pm = true;
+	/* Stmmac always requires an RX clock for hardware initialization */
+	priv->phylink_config.mac_requires_rxc = true;
 
 	phylink = phylink_create(&priv->phylink_config, fwnode,
 				 mode, &stmmac_phylink_mac_ops);
@@ -7842,16 +7844,12 @@ int stmmac_resume(struct device *dev)
 	}
 
 	rtnl_lock();
-	if (device_may_wakeup(priv->device) && priv->plat->pmt) {
-		phylink_resume(priv->phylink);
-	} else {
-		phylink_resume(priv->phylink);
-		if (device_may_wakeup(priv->device))
-			phylink_speed_up(priv->phylink);
-	}
-	rtnl_unlock();
 
-	rtnl_lock();
+	/* Prepare the PHY to resume, ensuring that its clocks which are
+	 * necessary for the MAC DMA reset to complete are running
+	 */
+	phylink_prepare_resume(priv->phylink);
+
 	mutex_lock(&priv->lock);
 
 	stmmac_reset_queues_param(priv);
@@ -7876,6 +7874,19 @@ int stmmac_resume(struct device *dev)
 	stmmac_enable_all_dma_irq(priv);
 
 	mutex_unlock(&priv->lock);
+
+	/* phylink_resume() must be called after the hardware has been
+	 * initialised because it may bring the link up immediately in a
+	 * workqueue thread, which will race with initialisation.
+	 */
+	if (device_may_wakeup(priv->device) && priv->plat->pmt) {
+		phylink_resume(priv->phylink);
+	} else {
+		phylink_resume(priv->phylink);
+		if (device_may_wakeup(priv->device))
+			phylink_speed_up(priv->phylink);
+	}
+
 	rtnl_unlock();
 
 	netif_device_attach(ndev);
--- a/drivers/net/phy/motorcomm.c
+++ b/drivers/net/phy/motorcomm.c
@@ -1168,6 +1168,25 @@ static int yt8531_config_init(struct phy_device *phydev)
 	return ret;
 }
 
+/*
+ * stmmac hard-resets the PHY on every resume (stmmac_mdio_reset() in
+ * stmmac_resume()) and, because it uses MAC-managed PM, phylib never
+ * calls phy_init_hw() again: the generic resume only clears BMCR_PDOWN and
+ * the extended registers stay at their power-on defaults (measured on an
+ * Orange Pi 5B: 0xa010 = 0x6bff, 0xa012 = 0xc8 after every rtcwake -m mem
+ * instead of what config_init wrote).  Redo the init before resuming.
+ */
+static int yt8531_resume(struct phy_device *phydev)
+{
+	int ret;
+
+	ret = yt8531_config_init(phydev);
+	if (ret < 0)
+		return ret;
+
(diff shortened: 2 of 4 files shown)


defcom5-rockchip:
CI note: the first "Build kernel on PR" run (34321240676) died on the runner itself, not in the build — the job annotation is `System.IO.DirectoryNotFoundException … _diag/blocks/… at GitHub.Runner.Common.PagingLogger.NewBlock()` on `cats-01`, with the arm64 build step still in progress and no build output. Re-pushed the same head (no content change) to trigger a fresh run. The three patches apply cleanly on the current `rk-6.1-rkr7.2` head and compile-test clean on an RK3588 config.

defcom5-rockchip:
Second run (34333006492, head 0fbad54 = same content) failed identically: `DirectoryNotFoundException … _diag/blocks/… at GitHub.Runner.Common.PagingLogger.NewBlock()` on `cats-01`, 5 minutes in, arm64 build step still in progress, no build output. Every other run of this workflow up to yesterday 15:07 UTC succeeded on the same runner, so this looks like the runner host (its `_diag/blocks` directory disappearing under the job) rather than the branch. I won't re-push again to avoid noise — a maintainer re-run whenever the runner is healthy should do it. For reference: the three patches apply cleanly on the current `rk-6.1-rkr7.2` head and the touched objects compile without warnings against an RK3588 config.

defcom5-rockchip:
Thank you, Igor. If rkr5.1 should get it as well, the same three patches apply there unchanged — say the word and I'll open the PR.

igorpecovnik:
rkr5.1 doesn't need it as we won't be adding any other fixes there or release it from that branch. We barely manage to maintain one branch.

PR #551, opened 2026-09-15
net: stmmac: don't claim IFF_UNICAST_FLT without address registers

rpardini:
dwmac4_setup() and dwmac1000_setup() set IFF_UNICAST_FLT unconditionally, and plat->unicast_filter_entries defaults to one whether or not the core was synthesized with additional MAC address registers.

On a core without them, dwmac4_set_filter() programs a secondary address into MAC_ADDRESS1, which does not exist, so nothing ever matches. The flag also stops __dev_set_rx_mode() from falling back to promiscuous mode, because it promised filtering the hardware cannot do, so the frames are dropped silently.

**RK3528** has such a core. Every DSA user port carries its own MAC, so dsa_slave_open() adds it to the master with dev_uc_add() and no user port can receive unicast; bringing up a second port happens to fix the first, by pushing netdev_uc_count() past unicast_filter_entries and into the promiscuous branch of dwmac4_set_filter().

Withdraw the flag when MAC_HW_FEATURE0.ADDMACADRSEL reports no additional registers, so the core falls back as it would for any MAC that cannot filter unicast.


defcom5-rockchip (changes requested):
The analysis matches the code and the fix does what it says for RK3528 — withdrawing the flag is the right way to let `__dev_set_rx_mode()` do the fallback. One thing before merge, and I measured it rather than guessed:

In rk-6.1, `GMAC_HW_FEAT_ADDMAC` is `BIT(18)` (`dwmac4.h:240`) and `dma_cap->multi_addr = (hw_cap & GMAC_HW_FEAT_ADDMAC) >> 18` (`dwmac4_dma.c:387`). `ADDMACADRSEL` in `MAC_HW_Feature0` is a 5-bit field, bits 22:18, so in this tree `multi_addr` is the low bit of the count, not the count. `!priv->dma_cap.multi_addr` is therefore true for 0 additional registers *and* for 2, 4, 6…30.

Measured on RK3588 (Orange Pi 5B, GMAC1 @ `0xfe1c0000`, `MAC_VERSION = 0x3051`, DWMAC 5.10a):

```
HW_FEATURE0 = 0x1a1173f7  ->  ADDMACADRSEL = 4   MACADR32SEL = 0   MACADR64SEL = 0
```

Four additional address registers, bit 18 zero. So `!priv->dma_cap.multi_addr` is true on RK3588, and this patch as written clears `IFF_UNICAST_FLT` on a core that has four working registers — any macvlan/bridge/DSA secondary address would push it to promiscuous instead of `MAC_ADDRESS1`, on every RK3588 in the tree.

Mainline fixed the parsing in 384ee2379a66 ("net: stmmac: dwmac4: Read the UC filter size from hardware capabilities", 2026-08-31, in torvalds/linux):

```
-#define GMAC_HW_FEAT_ADDMAC		BIT(18)
+#define GMAC_HW_FEAT_MACADR64SEL	BIT(24)
+#define GMAC_HW_FEAT_MACADR32SEL	BIT(23)
+#define GMAC_HW_FEAT_ADDMAC		GENMASK(22, 18)
```

Folding that hunk in ahead of this change makes the condition exact: RK3588 reads 4 and keeps its filter, RK3528 reads 0 and gets the fallback. Since the banks are independently selectable, `!multi_addr && !additional_32_addr && !additional_64_addr` would also cover a core that has only the 32/64 bank (the driver programs from register 1, so those would still miss).

Fuller option: 384ee2379a66 + 6815415d68b9 together with the net fixes "Account for primary MAC in UC filtering" (d29b399150b0..96e8cb5527ce) compute the real filter size from HW features and fall to PR inside `dwmac4_set_filter()` without touching the flag — RK3528 works, and RK3588 gets its five slots instead of the default one. Bigger backport; this patch plus the GENMASK hunk is fine for rkr7.2 if that isn't wanted yet.

Could you add RK3528's `HW_FEATURE0` value (bits 22:18, 23, 24) to the commit message so the "no additional registers" claim is on record? The dwmac1000 path is fine as-is — `DMA_HW_FEAT_ADDMAC` really is a single bit on the 3.x cores.

PR #556, opened 2026-09-20
arm64: dts: rockchip: fix the Mixtile Blade 3, Core3588E and Edge 2 device trees

evtest-hash:
## Summary

Fix the device tree descriptions for the Mixtile Blade 3, Core3588E and Edge 2
boards, including their overlays. One commit per board.

The changes were reviewed against the board schematics and the Rockchip vendor
kernel, and the affected boards were subsequently validated on hardware.

### Blade 3

- Rename `rk3588-blade3-v101-linux.dts` to `rk3588-mixtile-blade3.dts` and
  update the board identity to `Mixtile Blade 3` / `mixtile,blade3`.
- Correct the power tree: the board takes a 12-20V input and feeds the RK806
  and the three RK860x from the 4.0V buck. Drop the rails and the four
  `pcie*_avdd*` regulators the board does not have, add `vcc3v3_minipcie`.
- Fix the microSD regulator polarity and voltage constraints.
- Add the RKVENC and VOP supplies required for DVFS.
- Correct the suspend-mode configuration for the board's internal 32kHz source.
- Describe i2c3, i2c7 and i2c8; disable can2, which has no transceiver.
- Add the IMX415 camera overlay and metadata for the SATA2 overlay.

### Core3588E

- Correct the SD, USB 3.1 and DisplayPort descriptions.
- Correct the suspend configuration and the supplies that must stay up across
  suspend.
- Add the two IMX219 camera overlays for the CAM0 and CAM1 connectors.

### Edge 2

- Update the board compatible to `mixtile,edge2`.
- Replace the legacy Wi-Fi/Bluetooth platform-data nodes with the standard
  SDIO pwrseq and serdev bindings.
- Correct the Bluetooth LPO clock source: it comes from Y8, not RK809
  clkout2. This reverses 5aa56286b496.
- Remove unused audio interfaces and add headphone detection.
- Correct the M.2 lane select, eMMC HS400ES and USB OTG descriptions.
- Add the M.2 USB3 overlay.

## Validation

Blade 3, Core3588E and Edge 2 were each runtime-validated on their own
hardware after these device tree changes.

All three DTBs and the five Mixtile overlays build, and `git diff --check` is
clean. The two IMX219 overlays emit the same `reg_format` warnings as the
existing `radxa-nx5-io-rpi-camera-v2-cam1` overlay, which is inherent to
describing an I2C client inside a `/plugin/` fragment.

The Blade 3 rename changes the DTB name, so it needs the matching
`armbian/build` change, submitted separately as armbian/build#10774. That
one has to be merged after this.


rpardini (changes requested):
Thanks, Kevin.

Ref the Blade3 rename: that's going to cause breakage for existing users. We should keep the old name existing, with just an #include to the new name, and both in the Makefile. Similar to [this](https://github.com/armbian/linux-rockchip/blob/rk-6.1-rkr7.2/arch/arm64/boot/dts/rockchip/rk3588-nanopc-cm3588-nas.dts) - that way we can merge this independent of armbian/build, and old users keep on working.

rpardini (approved):

